PR #1, opened 2017-02-28
add build for minified js file

fanwgwg:
- Use `npm run build` to build source scripts `librarie.js` and `librarie.min.js`, output in `./dist/v0.0.1/`;
- By default `index.html` uses the minified script `librarie.min.js`;
- Upon version update, version in `webpack.config.js` and the script source of `librarie.min.js` in `index.html` needs to be updated.

@Benglin


package.json
```diff
@@ -21,9 +24,9 @@
     "@types/react": "^15.0.0",
     "@types/react-dom": "^0.14.20",
     "@types/node": "^7.0.5",
-    "@types/whatwg-fetch": "0.0.33",
     "awesome-typescript-loader": "^3.0.0-beta.18",
     "source-map-loader": "^0.1.5",
-    "typescript": "^2.1.4"
+    "typescript": "^2.1.4",
+    "uglify-js": "github:mishoo/UglifyJS2#harmony"
```
Benglin: Can we not add a dev dependency through NPM by doing this?

`npm install uglify-js --save-dev`

src/components/LibraryContainer.tsx
```diff
@@ -1,5 +1,4 @@
 /// <reference path="../../node_modules/@types/node/index.d.ts" />
-/// <reference path="../../node_modules/@types/whatwg-fetch/index.d.ts" />
```
Benglin: This file is updated, can you get the latest codes here?

index.html
```diff
@@ -24,7 +24,7 @@
         <script src="./node_modules/react-dom/dist/react-dom.js"></script>
 
         <!-- Main -->
-        <script src="./dist/bundle.js"></script>
+        <script src="./dist/librarie.min.js"></script>
```
Benglin: Is the `min` automatically added to `librarie.js`? And for the production build it will be just `librarie.js`?

package.json
fanwgwg: The default uglify-js installed via NPM only works on ES5, and they have one branch `harmony` for ES6.

index.html
fanwgwg: The `min` will only be created if we run `npm run build`, if we just run `webpack` or `npm run dev`, only librarie.js will be created, but I am not sure which one we should put here.

src/components/LibraryContainer.tsx
ramramps: Is it possible to create a config file, that includes all the dependencies? You can inject the config file at the start up either using require or the react or webpack way.

index.html
Benglin: Nice, okay please also update `ReadMe.md` to include these details.

src/components/LibraryContainer.tsx
Benglin: Thanks for the suggestion, @ramramps, do you have some examples of that? We can start looking into WebUI component for info if that's easier for you.

package.json
Benglin: Okay, can we give ES5 a try?

index.html
ramramps: you can put librarie.js instead of min. leave it to the consumers to decide.

src/components/LibraryContainer.tsx
ramramps: yes @Benglin. Flood / WebUIComponents have a config file.

index.html
fanwgwg: I've just updated so that webpack will update the script src itself based on the environment

README.md
```diff
@@ -53,8 +65,11 @@ The following simple HTML code illustrates the way to embed library view into an
         <script src="./node_modules/react/dist/react.js"></script>
         <script src="./node_modules/react-dom/dist/react-dom.js"></script>
 
-        <!-- The main library view component -->
-        <script src="./dist/bundle.js"></script>
+        <!-- The main library view compoment will be injected depending on build environment-->
+        <!-- start:librarie inject -->
+        <script src = './dist/librarie.min.js'></script>
+        <!-- end:librarie inject -->
+
```
Benglin: Can we restore these few lines? This is meant to be a reference document and therefore shouldn't include things like `inject` stuff.

package.json
```diff
@@ -18,12 +20,17 @@
...
     "awesome-typescript-loader": "^3.0.0-beta.18",
+    "babel-core": "^6.23.1",
+    "babel-loader": "^6.3.2",
+    "babel-preset-es2015": "^6.22.0",
+    "babel-preset-react": "^6.23.0",
+    "inject-html-webpack-plugin": "0.0.25",
     "source-map-loader": "^0.1.5",
-    "typescript": "^2.1.4"
+    "typescript": "^2.1.4",
+    "webpack": "^2.2.1"
```
Benglin: If we include both `typescript` and `webpack` as development dependencies, then you may want to update `ReadMe.md` to exclude the two steps for them with `npm install -g`.

README.md
```diff
@@ -6,15 +6,27 @@ Installing Webpack and TypeScript globally
...
+- Navigate to `localhost:3456` in Google Chrome browser
+
+### Easy production build
+- Build source scripts
+
+    `$ npm run build`
+
+- Serve up the new library view
+
+    `$ node ./index.js`
```
Benglin: Let's merge both `Easy production build` and `Easy development build` sections into one `Building librarie.js`. Then under *Build source scripts* we have:

    `$ npm run dev` (for development build)
    `$ npm run prod` (for production build)

Oh yes, change that `npm run build` to `npm run prod` for clarity.

package.json
```diff
@@ -18,12 +20,17 @@
...
+    "@types/node": "^7.0.5",
     "@types/react": "^15.0.0",
     "@types/react-dom": "^0.14.20",
-    "@types/node": "^7.0.5",
-    "@types/whatwg-fetch": "0.0.33",
     "awesome-typescript-loader": "^3.0.0-beta.18",
+    "babel-core": "^6.23.1",
+    "babel-loader": "^6.3.2",
+    "babel-preset-es2015": "^6.22.0",
+    "babel-preset-react": "^6.23.0",
```
Benglin: All these `babel` just to compile into ES5? Can `webpack` do it? I remember seeing some settings in `webpack.config` which tells what's the JavaScript version for output.

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "Project that contains all hosted contents of Dynamo Windows client",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "dev": "set NODE_ENV=0 && webpack --progress --colors",
+    "build": "set NODE_ENV=1 && webpack"
```
Benglin: Remember to rename `build` to `prod`. Also rename `NODE_ENV` to `production_build`, aligning better with the rest (also *node env* doesn't really mean anything).

webpack.config.js
```diff
@@ -1,15 +1,44 @@
+const webpack = require('webpack');
+var InjectHtmlPlugin = require('inject-html-webpack-plugin')
+var PROD = (process.env.NODE_ENV == 1);
```
Benglin: Perhaps rename `PROD` to `production_build` or `productionBuild`.

webpack.config.js
```diff
@@ -1,15 +1,44 @@
...
 module.exports = {
     entry: [
         "./src/LibraryUtilities.ts",
         "./src/LibraryView.tsx"
     ],
     target: "node",
     output: {
-        filename: "bundle.js",
+        filename: PROD ? "librarie.min.js" : "librarie.js",
         path: __dirname + "/dist",
```
Benglin: One thing though, we want to be releasing `librarie.js` with specific versions. Can we have a variable in this file that tells the version? Say, `let version = "0.0.1"` and then make this `path: __dirname + "/dist/" + version`?

Benglin: The final produced file would be `/dist/0.0.1/librarie.js`

README.md
fanwgwg: Is it okay if I just I just restore these few lines from `readme.md`, but keep them in `index.html`? Cuz currently my implementation in `webpack.config.js` requires these few lines to know where to inject the output js file.

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Project that contains all hosted contents of Dynamo Windows client",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "npm run dev & npm run prod",
+    "dev": "set production_build=0 & webpack",
+    "prod": "set production_build=1 & webpack"
```
Benglin: This is great, combining `dev` and `prod` into one `build` command, while still retain the original two so that one can choose to run either `dev` and `prod` separately.

README.md
```diff
@@ -2,19 +2,20 @@
...
+    $ npm install
 
-### Easy development build
+### Build and run librarie.js
 - Build source scripts
 
-    `$ webpack`
+    `$ npm run build`
+    
+    Two source scripts will be output to `./dist/v0.0.1/` folder. One is `librarie.js` for development build, another one is `librarie.min.js` for production build. By default `librarie.min.js` will be used in `index.html`.
```
Benglin: Perhaps rephrase it this way:

> Both `librarie.js` (development script) and `librarie.min.js` (production script) will be output in `./dist/v0.0.1/` folder. By default `librarie.min.js` will be referenced in `index.html`.

webpack.config.js
```diff
@@ -1,15 +1,33 @@
+const webpack = require('webpack');
+var productionBuild = (process.env.production_build == 1);
```
Benglin: This `var` should be `let`.

Benglin:
Thanks @fanwgwg, few more tweaks and we'll be good to go!

package.json
fanwgwg: ok just updated!

Benglin:
This is great, thanks for the great work, @fanwgwg! 👍 LGTM

PR #2, opened 2017-03-01
add some testcases

fanwgwg:
add some sample testcases with the dependencies set up


src/test/libraryUtilitiesTestClasses.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+
+export class layoutElementData {
+    text: any;
+    iconName: any;
+    elementType: any;
+    include: any;
+    childElements: layoutElementData[] = [];
+}
```
Benglin: We may want Pascal case for classes, and as much as possible, use type information. E.g. many of those `any` should really be `string`.

Benglin:
This pull request is superseded by [another pull request](https://github.com/DynamoDS/librarie.js/pull/8), closing it...

PR #3, opened 2017-03-08
remove LibraryItem class from LibraryUtilities

fanwgwg:
@Benglin


Benglin:
Thanks for such an awesome work, @fanwgwg! 👍

PR #5, opened 2017-03-09
Add search view

fanwgwg:
Add a search view that will display structured results while typing, based on substring search.

capture


src/LibraryUtilities.ts
```diff
@@ -249,3 +255,32 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
     }
     return convertToLibraryTree(typeListNodes, layoutElements);
 }
+
+function setItemVisible(item: ItemData) {
```
Benglin: Let's name this `showItemRecursive` instead, making the intents very clear.

src/LibraryUtilities.ts
```diff
@@ -249,3 +255,32 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+    item.visible = true;
+    item.childItems.forEach((childItem) => setItemVisible(childItem));
+}
+
+export function search(text: string, item: ItemData) {
+    let index = item.text.toLowerCase().indexOf(text);
```
Benglin: I think once `index` is set and it is not `-1`, then we can make the item visible and return, right? Otherwise the `for` loop below will overwrite its value.

src/LibraryUtilities.ts
```diff
@@ -249,3 +255,32 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+
+    for (let searchString of item.searchStrings) {
+        index = searchString.indexOf(text);
+        if (index >= 0) {
+            break;
+        }
+    }
+
+    if (index >= 0 && item.itemType !== "group") {
+        setItemVisible(item); // show all child items if text is found in parent item
```
Benglin: Do update the comments to be like so:

```js
if (index >= 0 && item.itemType !== "group") {
    // Show all items recursively if a given text is found in the current 
    // (parent) item. Note that this does not apply to items of "group" type
    setItemVisible(item);
}
```

src/LibraryUtilities.ts
Benglin: Shouldn't the above `indexOf` (and also the `for` loop below) to be done only for items that are not `group` type? If we can directly skip these steps for `group` items then it'll save time.

fanwgwg:
@Benglin I've just made some changes regarding your comment.

fanwgwg:
@Benglin Can merge this? Then I'll make another pull request for the list view

src/LibraryUtilities.ts
```diff
@@ -248,3 +254,39 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
 }
+
+function showItemRecursive(item: ItemData) {
+    item.visible = true;
+    item.childItems.forEach((childItem) => showItemRecursive(childItem));
+}
+
+export function search(text: string, item: ItemData) {
+    if (item.itemType !== "group") {
+        let index: number;
```
Benglin: Can we explicitly give `index` a value of `-1` here instead of leaving it unassigned? Because its value will be unknown if `item.searchStrings` loop below does not execute due to empty list.

src/components/LibraryItem.tsx
```diff
@@ -54,7 +57,21 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
+        };
+    }
+
+    componentWillReceiveProps(nextProps: LibraryItemProps) {
+        if (nextProps.data.visible !== this.state.visible) {
+            this.setState({ visible: nextProps.data.visible });
+        }
+        if (nextProps.data.expanded !== this.state.expanded) {
+            this.setState({ expanded: nextProps.data.expanded });
+        }
```
Benglin: Can you describe why is this `componentWillReceiveProps` needed (and what's the difference between this and the `constructor` above)? If this is needed, can we not merge both the `setState` into one? That way React.js doesn't have to queue up two state changes since they are aiming at the same component.

Benglin: I meant to suggest this:

```js
if ((nextProps.data.visible !== this.state.visible) || 
    (nextProps.data.expanded !== this.state.expanded)) {
        this.setState({
            visible: nextProps.data.visible,
            expanded: nextProps.data.expanded
        });
}
```

src/components/LibraryItem.tsx
```diff
@@ -73,31 +90,33 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         let nestedElements = null;
         let clusteredElements = null;
 
-        if (this.state.expanded) // Show only nested elements when expanded.
-        {
-            if (this.props.data.childItems && (this.props.data.childItems.length > 0)) {
+        // visible only nested elements when expanded.
+        if (this.state.expanded && this.state.visible && this.props.data.childItems.length > 0) {
```
Benglin: `this.props.data.childItems` condition is removed, is it always non-undefined?

Also this may look cleaner:

```js
let shouldDisplay = this.state.expanded && this.state.visible;
if (shouldDisplay && (this.props.data.childItems.length > 0)) {
    // ...
}
```

src/components/LibraryItem.tsx
```diff
@@ -73,31 +90,33 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
+                    {nestedElements}
                 </div>
-                {clusteredElements}
-                {nestedElements}
-            </div>
-        );
+            );
+        } else {
+            return null;
+        }
```
Benglin: Do we come all the way here (i.e. after computing all other stuff like `nestedElements`) just to `return null` when `this.state.visible === false`? Remember that will incur unnecessary calculations that can be avoided? If that's the case move `return null` all the way up (and remove other `if (this.state.visible)` related checks since it will be `true` once it passed the initial check.

src/components/SearchView.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import { LibraryItem } from "./LibraryItem";
+import { LibraryView } from "../LibraryView";
+import { search, ItemData } from "../LibraryUtilities";
+
+interface SearchViewProps {
+    onSearchChanged: any;
```
Benglin: This should be renamed to `onSearchModeChanged`, and it would be nice if you can [give it a type](https://www.typescriptlang.org/docs/handbook/interfaces.html) like so:

```js
interface SearchModeChangedFunc {
    (inSearchMode: boolean): void;
}

interface SearchViewProps {
    onSearchModeChanged: SearchModeChangedFunc;
}
```

src/components/SearchView.tsx
```diff
@@ -0,0 +1,81 @@
...
+    performSearch(text: string) {
+        this.props.items.forEach(item => search(text, item));
+    }
+
+    generateStructuredItems() {
+        let structuredItems: JSX.Element[] = [];
+        let index = 0;
+        structuredItems = this.props.items.map((item: ItemData) => <LibraryItem key={index++} libraryView={this.props.libraryView} data={item} />);
+        return structuredItems;
+    }
```
Benglin: Do clean it up this way:

```js
generateStructuredItems() {
    let index = 0;
    return this.props.items.map((item: ItemData) => 
        <LibraryItem key={index++} libraryView={this.props.libraryView} data={item} />);
}
```

src/components/SearchView.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    setItemVisibility(visibility: boolean, items: ItemData[] = this.props.items) {
+        items.forEach(item => {
+            item.visible = visibility;
+            this.setItemVisibility(visibility, item.childItems);
+        });
+    }
```
Benglin: Move `setItemVisibility` to utility class, I don't think `this` is required at all.

src/components/SearchView.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    setItemExpandability(expandability: boolean, items: ItemData[] = this.props.items) {
+        items.forEach(item => {
+            item.expanded = expandability;
+            this.setItemExpandability(expandability, item.childItems);
+        });
+    }
```
Benglin: I got the feeling that we are enabling/disabling the expandability here, I think a better term for `setItemExpandability` would `setItemExpandState` be nicer?

src/components/SearchView.tsx
```diff
@@ -0,0 +1,81 @@
...
+        super(props);
+
+        this.state = {
+            searchText: ''
+        };
+    }
+
+    performSearch(text: string) {
+        this.props.items.forEach(item => search(text, item));
+    }
```
Benglin: Move `performSearch` to utility class and add tests for it, maybe name it `searchItemRecursive`.

src/components/LibraryContainer.tsx
```diff
@@ -13,7 +14,9 @@ export interface LibraryContainerProps {
     layoutSpecsJson: any
 }
 
-export interface LibraryContainerStates { }
+export interface LibraryContainerStates {
+    inSearch: boolean
```
Benglin: I think `inSearchMode` can be a better name.

src/LibraryUtilities.ts
```diff
@@ -248,3 +254,39 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
     }
     return convertToLibraryTree(typeListNodes, layoutElements);
 }
+
+function showItemRecursive(item: ItemData) {
+    item.visible = true;
+    item.childItems.forEach((childItem) => showItemRecursive(childItem));
+}
```
Benglin: Move `showItemRecursive` method to utility class, then add test cases for it.

src/components/LibraryContainer.tsx
```diff
@@ -22,18 +25,38 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
 
         super(props);
+        this.state = { inSearch: false };
         this.generatedLibraryItems = buildLibraryItemsFromLayoutSpecs(
             this.props.loadedTypesJson, this.props.layoutSpecsJson);
     }
 
+    onSearchChanged(inSearch: boolean) {
+        this.setState({ inSearch: inSearch });
+    }
```
Benglin: Do rename this function to `onSearchModeChanged` too.

src/components/SearchView.tsx
Benglin: Move `setItemExpandability` to `LibraryUtilities.ts` and add test cases for it, I don't think `this` is required at all.

src/components/LibraryItem.tsx
fanwgwg: We will need this if we want to allow user to toggle the expansion in the structured search view. Since by default all the found items in search view have `expanded = true` stored in their `ItemData`, we also have a `expanded` state in the state of `LibraryItem`, this is needed to make sure that these two always the same. Otherwise there will be issues like which one should we check when rendering. However, if we just don't allow user to unexpand in search view, this will not be needed.

src/components/LibraryItem.tsx
fanwgwg: Actuallt in the `ItemData` class, we initialize `childItems: ItemData[] = []`, so it will not be undefined.

fanwgwg:
@Benglin Thanks a lot for the review! 
I've just updated the code accordingly, and added some testcases for those search-related functions in LibraryUtilities

src/LibraryUtilities.ts
```diff
@@ -245,3 +251,54 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+    }
+}
+
+export function showItemRecursive(item: ItemData) {
+    item.visible = true;
+    item.expanded = true;
+    for(let childItem of item.childItems) {
+        showItemRecursive(childItem);
+    }
+}
```
Benglin: Hi @fanwgwg, thanks for doing this. In a subsequent pull request you can merge `resetItemData` and `showItemRecursive` into one that looks like this:

```js
export function setItemStateRecursive(items: ItemData[], visible: boolean, expanded: boolean) {
    // Makes use of visible and expanded parameters
}
```

src/components/LibraryContainer.tsx
```diff
@@ -22,18 +25,37 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
     }
 
+    onSearchModeChanged(inSearchMode: boolean) {
+        this.setState({ inSearchMode: inSearchMode });
+    }
+
     render() {
 
         try {
             let index = 0;
```
Benglin: You can do this later, but let's move `index` into the `else` clause below before the `map`.

src/components/LibraryItem.tsx
```diff
@@ -54,10 +60,23 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
+        // Assign initial state
+        this.state = {
+            expanded: this.props.data.expanded
+        };
+    }
+
+    componentWillReceiveProps(nextProps: LibraryItemProps) {
+        if (nextProps.data.expanded !== this.state.expanded) {
+            this.setState({ expanded: nextProps.data.expanded });
+        }
```
Benglin: Please add some comments to describe why is this needed in a subsequent PR.

src/components/SearchView.tsx
```diff
@@ -0,0 +1,66 @@
...
+            listItems = this.generateStructuredItems();
+        } else {
+            // Reset ItemData when search text is cleared
+            resetItemData(this.props.items);
+        }
+
+        return (
+            <div className="searchView">
+                <div className="searchBar">
+                    <input id="searchInput" type="search" placeholder="Search..." onChange={this.onTextChange.bind(this)}></input>
```
Benglin: Hi @fanwgwg, can you make a task for this? We want `onChange` to not trigger a search so quickly, it should start a timer of maybe `300ms` and if that time expired then we will trigger the search. Otherwise if a new character is being pressed, then the timer will restart to be `300ms`. Thanks!

fanwgwg: ok sure! I've just made some changes according to the preivous comments.

Benglin:
Thanks again for being so innovative on the search view, we didn't think of that being possible before so good job there! Please follow up with other comments in this PR later.

PR #7, opened 2017-03-10
Function callback mechanism

yeexinc:
[(Reference)](http://stackoverflow.com/questions/15308371/custom-events-model-without-using-dom-events-in-javascript)

Several items are added:
1. `Event` class which stores a number of callback functions together with a name (string) as an identifier.
2. `Reactor` class which stores a number of events.
-----
To register an event, add a new callback function and specify a name for it:
```
libView.on("somethingHappened", function() {
            // do something
        })
```

To fire the function, use the `raiseEvent()` function, and pass in the identifier (name):
```
libView.raiseEvent("somethingHappened");
```

To register a callback function that takes in a parameter, initialize the function and specify the parameter. While dispatching the event, pass in the input as the second parameter of `raiseEvent()`.
```
libView.on("callbackFuncWithParam", function(paramName: string) {
    console.log(x);
})
```
```
libView.raiseEvent("callbackFuncWithParam", "string parameter");
// (result:) string parameter
```
Note: Callback functions are limited to taking in 0 or 1 parameters only.

-----

Please take a look and if this implementation is okay, I will update the readme file for it.


src/components/EventHandler.tsx
```diff
@@ -0,0 +1,65 @@
+
```
Benglin: This file (and the next one `LibraryEntryPoint.tsx` are not *React.js components* so they should not be under `/src/components` folder, they should just be under `/src` folder. Also, since they do not contain any JSX/TSX contents, they should just be having an extension of `*.ts`, not `*.tsx`.

Benglin: Also, @fanwgwg has suggested a nice thing about auto-formatting our `librarie.js` codes, can you try it out? *VS Code* has a nice built in way of auto-formatting the current source file, on Mac it is `Shift + Option + F` not sure what is it on Windows.

src/LibraryView.tsx
```diff
@@ -17,13 +18,16 @@ export class LibraryView {
...
     layoutSpecsJson: any = null;
+    reactor: Reactor = null;
 
     constructor(config: LibraryViewConfig) {
 
         this.setLoadedTypesJson = this.setLoadedTypesJson.bind(this);
         this.setLayoutSpecsJson = this.setLayoutSpecsJson.bind(this);
         this.prefetchContents = this.prefetchContents.bind(this);
         this.updateContentsInternal = this.updateContentsInternal.bind(this);
+        LibraryView.prototype.reactor = new Reactor();
```
Benglin: Since we're in TypeScript, let's not use `prototype` anymore. It can just be `this.reactor = new Reactor();`.

src/components/EventHandler.tsx
```diff
@@ -0,0 +1,65 @@
...
+    constructor(name: string) {
+        this.name = name;
+        this.callback = null;
+    }
+
+    registerCallback(callback: Function) {
+        this.callback = callback;
+    }
+}
```
Benglin: This doesn't seem to be aligned with the referenced document. One `Event` has one `name`, and one or more `callback` functions. The `Reactor` is meant to contain a list of `Event`, each `Event` represents one event with a given string. For example if we have three callbacks for `itemClicked` and two callbacks for `itemExpanded`, then we will only have two `Event` inside `Reactor` (i.e. one for `itemClicked` and another for `itemExpanded`).

Benglin: And since `Event` has multiple callbacks, then it should contain a method like this:

```ts
export class Event {
    raiseEvent(params?: any | any[]) {
        this.callbacks.forEach(...);
    }
}
```

src/components/LibraryItem.tsx
```diff
@@ -184,9 +185,6 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     }
 
     onLibraryItemClicked() {
-
-        // Toggle expansion state.
-        let currentlyExpanded = this.state.expanded;
-        this.setState({ expanded: !currentlyExpanded });
+        LibraryEntryPoint.dispatchEvent("libraryItemClick", this);
```
Benglin: Why are we moving the stuff that is for `LibraryItem` out of it? This is what is expected:

```ts
onLibraryItemClicked() {
    this.setState({ expanded: !currentlyExpanded });

    // I think we need a global instance of Reactor object
    reactor.raiseEvent("itemClicked", this. data.creationName);
}

```

src/LibraryView.tsx
```diff
@@ -17,13 +18,16 @@ export class LibraryView {
...
+        registerLibraryEvents();
```
Benglin: Why do we need to explicitly call this here? I think you mistaken the intent here, `LibraryView` itself does not need to register its own handler, see `onLibraryItemClicked()` comment below.

src/LibraryView.tsx
Benglin: After reviewing further, I think we need to pass `LibraryView` to all the sub-items (i.e. `LibraryItem` and `ClusterView`) so that those classes can have access to this copy of the `Reactor`. For example, in all the places that we need to raise an event, the following should be done:

```ts
// Now we also need to pass LibraryView to LibraryItem through its property
export interface LibraryItemProps { libraryView: LibraryView, data: ItemData }

export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemState> {

  onSomethingHappened() {
    let libraryView = this.props.libraryView; // Then we obtain the LibraryView like this
    libraryView.raiseEvent("somethingHappened", "some params");
}
```

Benglin: Let me make that PR, hang on.

src/components/LibraryItem.tsx
Benglin: We will get the `LibraryView` instance from here and then get it to raise the event instead of accessing the `Reactor` directly like this. Let me make a PR.

src/components/EventHandler.tsx
```diff
@@ -0,0 +1,65 @@
...
+                        console.log(e);
+                    }
+                }
+            })
+        }
+    }
+}
+
+// This function is called in LibraryView when the Reactor is initialized.
+export function registerLibraryEvents(): void {
```
Benglin: Please remove this function it is not needed.

src/components/LibraryEntryPoint.tsx
```diff
@@ -0,0 +1,14 @@
+
```
Benglin: Please remove this file it isn't needed.

src/components/LibraryItem.tsx
Benglin: Okay here's how the final code should look like:

```ts
onLibraryItemClicked() {
    this.setState({ expanded: !currentlyExpanded });

    let libraryView = this.props.libraryView;
    libraryView.raiseEvent("itemClicked", this. data.creationName);
}
```

And in the `LibraryView`, the code should look this way:

```ts
export class LibraryView {

    reactor: Reactor = null;

    constructor() {
        this.reactor = new Reactor();
    }

    on(eventName: string, callback: Function) {
        this.reactor.registerEvent(eventName, callback);
    }

    raiseEvent(params?: any | any[]): void {
        this.reactor.raiseEvent(params);
    }
}
```

Then how an HTML site registers a callback is like this:

```ts
let libraryView = ...;
libraryView.on("itemClicked", function(creationName) {
    console.log(creationName); // This should output creation name of the clicked item.
});
```

yeexinc: @Benglin Hi Ben, thank you very much for the detailed review.
May I know where should the libraryView.on function be called to register the callback?

Benglin: We'll leave that to the `index.html` or any other clients who are using `librarie.js`

Benglin:
Thanks again @yeexinc! Can you update the `ReadMe.md` file to include the event part where the HTML calls `libView.on("itemClicked", ...)` part?

Benglin:
Also the description of this PR is off, need updating. Thanks!

yeexinc:
@Benglin I was thinking of accessing the `Event` object this way so that if the returned index is not `-1`, it can register the callback:
`this.events[index].registerCallback(callback);`
Is this good?

README.md
```diff
@@ -68,11 +68,39 @@ The following simple HTML code illustrates the way to embed library view into an
 
             let libView = new LibraryEntryPoint.LibraryView(configuration);
 
-            // TODO: The callback when a library item is clicked on has not been 
-            // provided yet, it is scheduled to be done in the near future.
+            // Register callback functions here.
         </script>
 
     </body>
 </html>
```
Benglin: Thanks for the write-up, @yeexinc. Unfortunately none of the following will be required, as a consumer of `librarie.js`, one will not be concerned about how events are raised -- they only care about how the events can be registered. Please remove the changes below in `ReadMe.md`.

README.md
```diff
@@ -68,11 +68,39 @@ The following simple HTML code illustrates the way to embed library view into an
 
             let libView = new LibraryEntryPoint.LibraryView(configuration);
 
-            // TODO: The callback when a library item is clicked on has not been 
-            // provided yet, it is scheduled to be done in the near future.
+            // Register callback functions here.
```
Benglin: This should be updated:

```js
libView.on("itemClicked", function(data) {
    console.log(data);
})
```

That is all we need for `ReadMe.md`.

index.html
```diff
@@ -36,6 +36,10 @@
         };
 
         let libView = new LibraryEntryPoint.LibraryView(configuration);
+
+        libView.on("itemClicked", function(item) {
+            console.log(item.props.data.text);
+        })
```
Benglin: The client doesn't care about the props, they just need the string. This is incorrect. Just do `console.log(item)` and update the code below. No one else outside of `librarie.js` will know the internals of `item.props.data`...

src/components/LibraryItem.tsx
```diff
@@ -188,9 +188,10 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     }
 
     onLibraryItemClicked() {
-
-        // Toggle expansion state.
         let currentlyExpanded = this.state.expanded;
         this.setState({ expanded: !currentlyExpanded });
+
+        let libraryView = this.props.libraryView;
+        libraryView.raiseEvent("itemClicked", this);
```
Benglin: Change it to `libraryView.raiseEvent("itemClicked", this.item.props.data. creationName);`

src/LibraryView.tsx
```diff
@@ -72,4 +75,12 @@ export class LibraryView {
             loadedTypesJson={this.loadedTypesJson}
             layoutSpecsJson={this.layoutSpecsJson} />, htmlElement);
     }
+
+    on(eventName: string, callback: Function) {
+        this.reactor.registerEvent(eventName, callback);
+    }
```
Benglin: Please check with @fanwgwg to see if this `on` method will be mangled (i.e. uglified). If it will be mangled into some other names, then external clients cannot rely on it, get help from Wei Guang to avoid the mangling to be applied on this `on` method. The `raiseEvent` below is internally used so it doesn't matter if it is mangled, leave that as it is.

Benglin:
Hi @yeexinc, don't return an `index` (and do another array element look-up), return the actual `Event` will be better, just check for `null` or use the `Event` if one is returned.

src/LibraryView.tsx
fanwgwg: okay I can manully set it to reserve the `on` keyword

yeexinc:
Thanks @Benglin , changes have been made accordingly.

package.json
```diff
@@ -16,9 +16,11 @@
   "author": "Dynamo Team",
   "license": "MIT",
   "dependencies": {
+    "@types/underscore": "^1.8.0",
```
Benglin: This should really be `devDependencies` because we don't ship it to customers.

src/EventHandler.ts
```diff
@@ -0,0 +1,68 @@
...
+    registerEvent(eventName: string, callback: Function) {
+        let event = this.getEvent(eventName);
+        if (event == null) {
+            event = new Event(eventName);
+            event.registerCallback(callback);
+            this.events.push(event);
+        }
+        else {
+            event.registerCallback(callback);
+        }
```
Benglin: A better way would be this:

```js
let event = this.getEvent(eventName);
if (!event) {
    event = new Event(eventName);
    this.events.push(event);
}

event.registerCallback(callback);
```

Also, checking with `!event` is better IMHO because it also covers the case where `undefined` is being returned.

src/components/LibraryItem.tsx
```diff
@@ -192,5 +192,8 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         // Toggle expansion state.
         let currentlyExpanded = this.state.expanded;
         this.setState({ expanded: !currentlyExpanded });
+
+        let libraryView = this.props.libraryView;
+        libraryView.raiseEvent("itemClicked", this.props.data.text);
```
Benglin: As mentioned in my previous comment, this should be `this.props.data.creationName`.

README.md
```diff
@@ -67,12 +67,18 @@ The following simple HTML code illustrates the way to embed library view into an
             };
 
             let libView = new LibraryEntryPoint.LibraryView(configuration);
-
-            // TODO: The callback when a library item is clicked on has not been 
-            // provided yet, it is scheduled to be done in the near future.
```
Benglin: We need the code here as well:

```js
libView.on("itemClicked", function(data) {
    console.log(data);
});
```

yeexinc:
@Benglin Thanks for the review, `creationName` was previously unassigned, so I updated `constructFromLayoutElement` for that.

src/LibraryView.tsx
Benglin: Thanks for the tips, @fanwgwg!

Benglin:
Thanks @yeexinc for the work 👍

PR #10, opened 2017-03-17
Support absolute url for icon resources

yeexinc:
This is to address [DYN-636](https://jira.autodesk.com/browse/DYN-636).
`creationName` is renamed to `contextData` in `RawTypeData.json`. Furthermore, `iconName`s are updated to include absolute urls of the icons.

Also, I was thinking if we could have a `pathIcon` property that specifies an icon for a `path` as well.

e.g.
```
...
"include": [
             {
              "path": "Autodesk.DesignScript.Geometry.BoundingBox",
              "pathIcon": "/src/resources/icons/Autodesk.DesignScript.Geometry.BoundingBox.png"
             },
...
```

Can render the icon for BoundingBox, and at the same time include the nodes which have `Autodesk.DesignScript.Geometry.BoundingBox` in their names:

capture

This is for expandable library items whose element type is `none`. Thoughts?

@Benglin

### DIFF
--- a/docs/LayoutSpecs.json
+++ b/docs/LayoutSpecs.json
@@ -2,7 +2,7 @@
   "elements": [
     {
       "text": "List",
-      "iconName": "Category.List.svg",
+      "iconName": "/src/resources/icons/Category.List.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
@@ -217,7 +217,7 @@
           "childElements": [
             {
               "text": "Keys",
-              "iconName": "Category.Keys.svg",
+              "iconName": "/src/resources/icons/Category.Keys.svg",
               "elementType": "category",
               "include": [
                 {
@@ -311,7 +311,7 @@
     },
     {
       "text": "Input",
-      "iconName": "Category.Input.svg",
+      "iconName": "/src/resources/icons/Category.Input.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
@@ -375,7 +375,8 @@
               "path": "DSCore.DateTime.TimeOfDay"
             },
             {
-              "path": "DSCore.DayOfWeek"
+              "path": "DSCore.DayOfWeek",
+              "pathIcon": "/src/resources/icons/DSCore.DayOfWeek.png"
             }
           ],
           "childElements": []
@@ -520,7 +521,7 @@
     },
     {
       "text": "Math",
-      "iconName": "Category.Math.svg",
+      "iconName": "/src/resources/icons/Category.Math.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
@@ -695,7 +696,7 @@
         },
         {
           "text": "Operators",
-          "iconName": "Category.Operators.svg",
+          "iconName": "/src/resources/icons/Category.Operators.svg",
           "elementType": "group",
           "include": [
             {
@@ -764,7 +765,7 @@
     },
     {
       "text": "Script",
-      "iconName": "Category.Script.svg",
+      "iconName": "/src/resources/icons/Category.Script.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
@@ -820,7 +821,7 @@
     },
     {
       "text": "Display",
-      "iconName": "Category.Display.svg",
+      "iconName": "/src/resources/icons/Category.Display.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
@@ -906,7 +907,7 @@
     },
     {
       "text": "ImportExport",
-      "iconName": "Category.ImportExport.svg",
+      "iconName": "/src/resources/icons/Category.ImportExport.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
@@ -1042,7 +1043,7 @@
     },
     {
       "text": "String",
-      "iconName": "Category.String.svg",
+      "iconName": "/src/resources/icons/Category.String.svg",
       "elementType": "category",
       "include": [],
       "childElements": [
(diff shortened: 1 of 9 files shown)


docs/LayoutSpecs.json
```diff
@@ -2,7 +2,7 @@
   "elements": [
     {
       "text": "List",
-      "iconName": "Category.List.svg",
+      "iconName": "/src/resources/icons/Category.List.svg",
```
Benglin: Thanks @yeexinc, since we're renaming stuff (i.e. breaking the contract), can we rename `iconName` to `iconUrl`?

docs/LayoutSpecs.json
```diff
@@ -375,7 +375,8 @@
               "path": "DSCore.DateTime.TimeOfDay"
             },
             {
-              "path": "DSCore.DayOfWeek"
+              "path": "DSCore.DayOfWeek",
+              "pathIcon": "/src/resources/icons/DSCore.DayOfWeek.png"
```
Benglin: Hi @yeexinc, do you mean if we don't specify `pathIcon` here, then no icon will be rendered for `DSCore.DayOfWeek` on the library view? I'm trying to see if `pathIcon` is absolutely necessary. I think we need it here because the location of the icons need to be absolute path, correct me if I'm wrong. In that case, can we rename `pathIcon` to `iconUrl` just to be consistent?

docs/LayoutSpecs.json
```diff
@@ -375,7 +375,8 @@
               "path": "DSCore.DateTime.TimeOfDay"
```
Benglin: Will we have issues with `DSCore.DateTime.TimeOfDay` not having an `iconUrl` value? Does it show the icons now?

src/LibraryUtilities.ts
```diff
@@ -12,13 +12,14 @@ export class TypeListNode {
     constructor(data: any) {
         this.fullyQualifiedName = data.fullyQualifiedName;
         this.iconName = data.iconName;
-        this.creationName = data.creationName;
+        this.creationName = data.contextData;
```
Benglin: Is there a reason we don't rename `creationName`?

Benglin:
Thanks @yeexinc, please remember to update both [loaded-data-types.md](https://github.com/DynamoDS/librarie.js/blob/master/docs/v0.0.1/loaded-data-types.md) and [layout-specs.md](https://github.com/DynamoDS/librarie.js/blob/master/docs/v0.0.1/layout-specs.md) with correct contents (please also enhance the description if it is affected).

docs/LayoutSpecs.json
yeexinc: @Benglin If the icon url is unspecified, it is constructed using the name of the path. Since we're not adding 'src/resources/icons/' and '.png' extensions anymore, the resulting icon url is http://localhost:3456/DSCore.DayOfWeek. This only applies to nodes in the 'none' category.

docs/LayoutSpecs.json
yeexinc: `DSCore.DateTime.TimeOfDay` is a leaf node (not in the `none` category), so its iconUrl is obtained from RawTypeData.json:
```
{
      "fullyQualifiedName": "DSCore.DateTime.TimeOfDay",
      "iconUrl": "/src/resources/icons/DSCore.DateTime.TimeOfDay.png",
      "contextData": "DSCore.DateTime.TimeOfDay@var",
      "itemType": "action"
},
```

yeexinc:
Thanks @Benglin for the review, I've updated the files accordingly.

docs/v0.0.1/loaded-data-types.md
```diff
@@ -6,9 +6,9 @@ This flat list of loaded data types takes the following form: it must contain ex
 
 - `fullyQualifiedName` - Fully qualified name of a given type (e.g. a method or property in a class). An example of this would be `Autodesk.DesignScript.Geometry.Curve.Offset` property. The fully qualified name is used for inclusion/exclusion under a given branch of the library after transformation (transformation is done through a [Layout Specification](./layout-specs.md) document).
 
-- `iconName` - Icon name of a given type. This can optionally include `*.svg` extension; if it does not contain `*.svg` extension, then a `.png` is appended to the icon name by default. **Note**: absolute URL of icon is not currently supported yet.
+- `iconUrl` - Absolute URL of icon of a given type.
```
Benglin: Thanks for updating the relevant descriptions!

src/LibraryUtilities.ts
```diff
@@ -103,17 +105,12 @@ export function constructNestedLibraryItems(
     for (let i = startIndex; i < fullNameParts.length; i++) {
         let libraryItem = new ItemData(fullNameParts[i]);
         libraryItem.itemType = "none";
-
-        // If 'i' is now '2' (i.e. it points to 'C'), then we will construct 
-        // the iconName as 'A.B.C'. And since the second parameter of 'slice' 
-        // is exclusive, we add '1' to it otherwise 'C' won't be included.
-        //  
-        libraryItem.iconName = fullNameParts.slice(0, i + 1).join(".");
+        libraryItem.iconUrl = iconUrl;
```
Benglin: We can keep the original logic and use it just in case `iconUrl` is not provided:

```js
if (iconUrl) {
    libraryItem.iconUrl = iconUrl;
} else {
    // If 'i' is now '2' (i.e. it points to 'C'), then we will construct 
    // ...
    libraryItem.iconName = fullNameParts.slice(0, i + 1).join(".");
}
```

Benglin:
Thanks @yeexinc for the hard work, I'm gonna merge this in for now, and you can fix the two requests after this 👍

PR #11, opened 2017-03-21
Add lines to indicate indentation

fanwgwg:
UI screenshot as below.
* For level 0, there would be no indentation.
* For level >= 1, `|--` would only appear at last level. 
* Upon expansion, indentation would change from `|--` to `|__`

***New screenshots are in comment below***

capture
capture2


src/components/ClusterView.tsx
```diff
@@ -53,7 +53,7 @@ export class ClusterView extends React.Component<ClusterViewProps, undefined> {
 
         let index = 0;
         return this.props.childItems.map((item: ItemData) => {
-            return (<LibraryItem key={index++} libraryView={this.props.libraryView} data={item} />);
+            return (<LibraryItem key={index++} libraryView={this.props.libraryView} data={item} level={0} />);
```
Benglin: Can we use `indentation` or `indentLevel` instead of `level`?

src/resources/ui/indentation1.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0" y="0" width="40" height="40" viewBox="0 0 40 40" enable-background="new 0 0 40 40" xml:space="preserve">
+  <g stroke="#aaaaaa"  stroke-width="1" width="40" height="40">
+    <line x1="20" y1="0" x2="20" y2="40"/>
+  </g>
+</svg>
```
Benglin: Can we rename these three files to `indent-line-i.svg`, `t.svg` and `l.svg`, respectively?

fanwgwg:
@Benglin Just updated the renamings! Regarding the indentation look on ui, is this okay?

fanwgwg:
@Benglin Screenshot with lines are added here

capture
capture1

Benglin:
Looks great, @fanwgwg. If you can address the merge conflict I'll merge this in.

fanwgwg:
Ok just resolved the conflicts with master @Benglin

src/components/LibraryItem.tsx
```diff
@@ -126,6 +109,23 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
+                    indentationPath += clusteredElements ? "indent-line-l.svg" : "indent-line-t.svg";
+                } else {
+                    indentationPath += "indent-line-i.svg";
+                }
+                indents.push(<img key={i} className={"Indentation"} src={indentationPath} />);
+            }
+
+            indentation = (<div className={"Indents"}>{indents}</div>);
+        }
+
```
Benglin: Can you help us reduce the above lines to:

```js
let indentElements = this.getIndentElements();
```

The `render()` function gets a little too long now.

Benglin: And of course I think we should update all methods that return elements to have a return type of `JSX.Element` or something along that line.

fanwgwg:
The method for indentation is now extracted out from render. @Benglin

src/components/LibraryItem.tsx
```diff
@@ -118,6 +122,29 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         );
     }
 
+    getIndentElements() {
+        let indentElements = null;
```
Benglin: Hi @fanwgwg, this method is still not having a return type of `JSX.Element`, it should. The same applies to `getNestedElements` and `getClusteredElements`. We need to be specific about the types in TypeScript.

Also, the method contents can be refactored this way:

```js
getIndentElements(): JSX.Element {
    if (this.props.indentLevel <= 0) {
        return null;
    }

    let indents: JSX.Element[] = []; // Being explicitly typed
    let indentationIconRootPath = "/src/resources/ui/"; // Doesn't need type because it's clear from assignment

    // Other stuff ...
}
```

Benglin:
Thanks @fanwgwg, this is fantastic! Demo time!

PR #12, opened 2017-03-21
Add documentation for search

fanwgwg:
- Add documentation for search item json format
- Add search entry point as `queryString` in `index.html`

### DIFF
--- a/docs/v0.0.1/search-items.md
+++ b/docs/v0.0.1/search-items.md
@@ -1,13 +1,13 @@
-# Search Item Type
-Search view consumes a flat list of search item type (in `json` format) for its contents.
+# Search Items
+Search view consumes a flat list of search items (in `json` format) for its contents.
 
-## Search item type JSON format
-This flat list of search item type takes the following form: it must contain exactly one `searchItemTypes` root item that represents a list of search item types. Each of the search item types is made up of the following fields: 
+## Search item JSON format
+This flat list of search items takes the following form: it must contain exactly one `searchItems` root item that represents a list of search items. Each of the search items is made up of the following fields: 
 
 - `text` - The name of a given search item.
 - `iconUrl`- The absolute URL of icon of a given search item.
 - `contextData` - The string value that will be passed to the event handler when a given search item is clicked.
-- `itemType` - This value can either be creation, action or query. For more details of these values, please refer to [Layout Specification](./layout-specs.md) document.
+- `itemType` - This value can either be `creation`, `action` or `query`. For more details of these values, please refer to [Layout Specification](./layout-specs.md) document.
 - `group` - The name of the group that a given search item belongs to.
 - `category` - The name of the category that a given search item belongs to.
 - `description` - The description text of the given search item.
@@ -15,7 +15,7 @@ This flat list of search item type takes the following form: it must contain exa
 
 ```json
 {
-  "searchItemTypes": [
+  "searchItems": [
     {
       "text": "ByCoordinates",
       "iconUrl": "/src/resources/icons/Autodesk.DesignScript.Geometry.Point.ByCoordinates.double-double.png",
--- a/docs/v0.0.1/search-items.md
+++ b/docs/v0.0.1/search-items.md
@@ -8,8 +8,6 @@ This flat list of search result items takes the following form: it must contain 
 - `iconUrl`- The absolute URL of icon of a given search result item.
 - `contextData` - The string value that will be passed to the event handler when a given search result item is clicked.
 - `itemType` - This value can either be `creation`, `action` or `query`. For more details of these values, please refer to [Layout Specification](./layout-specs.md) document.
-- `group` - The name of the group that a given search result item belongs to.
-- `category` - The name of the category that a given search result item belongs to.
 - `description` - The description text of a given search result item.
 - More fields will be added soon.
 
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -32,7 +32,8 @@
         let configuration = {
             htmlElementId: "libraryContainerPlaceholder",
             loadedTypesUrl: "loadedTypes", // A url where loaded types are to be downloaded.
-            layoutSpecsUrl: "layoutSpecs"  // A url where layout specs are to be downloaded.
+            layoutSpecsUrl: "layoutSpecs",  // A url where layout specs are to be downloaded.
+            queryString: ""
```
Benglin: This should be `searchEndpoint` and not `queryString`.

src/LibraryView.tsx
```diff
@@ -61,6 +63,17 @@ export class LibraryView {
...
+        // fetch search items (in json format) if queryString exists and is not empty
+        if (queryString && queryString.length > 0) {
+            fetch(queryString)
+                .then(function (response: Response) {
+                    return response.text();
+                }).then(function (jsonString) {
+                    thisObject.searchItemsJson = JSON.parse(jsonString);
+                    thisObject.updateContentsInternal();
+                });
+        }
```
Benglin: This doesn't belong here, I think you have misunderstood the purpose of `searchEndpoint`, it should be a fully qualified URL to a search REST API. E.g. `https://dynamoreach.com/search-library` and when user types in search keyword, we will then call it like so:

```
https://dynamoreach.com/search-library?query=PointByCoordinates
```

Does that make sense?

ramramps: Hey @Benglin we don't have http endpoint for REACH.  You should use  Command /Message pattern for the requests.

src/LibraryView.tsx
```diff
@@ -61,6 +63,17 @@ export class LibraryView {
                 thisObject.layoutSpecsJson = JSON.parse(jsonString);
                 thisObject.updateContentsInternal();
             });
+
+        // fetch search items (in json format) if queryString exists and is not empty
+        if (queryString && queryString.length > 0) {
+            fetch(queryString)
```
ramramps: FETCH is awesome API. But it doesn't work on all the browsers.. just add a comment here with browser compatibility.

ramramps:
@Benglin As we discussed, the URL is not a URL. LibraryView should use the DATA directly, instead of fetching from a URL. IOW, the config takes a JSON object and LibraryView should not worry about how/where that data is fetched.  Also, the queryString is awesome, but that offsets the component. May be I don't understand the querystring needs.

docs/v0.0.1/search-items.md
```diff
@@ -0,0 +1,41 @@
...
+Search view consumes a flat list of search items (in `json` format) for its contents.
+
+## Search item JSON format
+This flat list of search items takes the following form: it must contain exactly one `searchItems` root item that represents a list of search items. Each of the search items is made up of the following fields: 
+
+- `text` - The name of a given search item.
+- `iconUrl`- The absolute URL of icon of a given search item.
+- `contextData` - The string value that will be passed to the event handler when a given search item is clicked.
+- `itemType` - This value can either be `creation`, `action` or `query`. For more details of these values, please refer to [Layout Specification](./layout-specs.md) document.
+- `group` - The name of the group that a given search item belongs to.
```
Benglin: I think we don't need `group` in search results, do we?

docs/v0.0.1/search-items.md
```diff
@@ -0,0 +1,41 @@
+# Search Items
```
Benglin: Can we rename it to `Search Result Item` just to be more descriptive?

docs/v0.0.1/search-items.md
fanwgwg: I think this will depend on what we want to show in the list mode. Cuz currently in dynamo it shows like `[group name][item type icon][category name]`

fanwgwg:
I've removed the queryString from this PR, now this PR is only for the documentation. I will create another PR to add a search callback function and extract the download to `index.html`.

Benglin:
Yes @ramramps I have already conveyed the idea to my team and it will be worked on soon. `librarie.js` will only handle incoming JSON, regardless of where it comes from. Thanks for the tips!

docs/v0.0.1/search-items.md
```diff
@@ -0,0 +1,41 @@
...
+Search view consumes a flat list of search result items (in `json` format) for its contents.
+
+## Search item JSON format
+This flat list of search result items takes the following form: it must contain exactly one `searchItems` root item that represents a list of search result items. Each of the search result items is made up of the following fields: 
+
+- `text` - The name of a given search result item.
+- `iconUrl`- The absolute URL of icon of a given search result item.
+- `contextData` - The string value that will be passed to the event handler when a given search result item is clicked.
+- `itemType` - This value can either be `creation`, `action` or `query`. For more details of these values, please refer to [Layout Specification](./layout-specs.md) document.
+- `group` - The name of the group that a given search result item belongs to.
```
Benglin: I remember recommending against having `group` in my previous review, do you have a different opinion on that?

src/LibraryView.tsx
```diff
@@ -70,10 +70,13 @@ export class LibraryView {
         }
```
Benglin: Please revert this file and I'll merge the PR. Thanks!

PR #13, opened 2017-03-22
Check for left-over items in library view

yeexinc:
This PR adds extra checking for processing the nodes in `LayoutSpecs.json` and `RawTypeData.json`:
- Nodes that have the same names but different parameters are included by specifying the path once, e.g. `DSCore.Math.Random@int` and `DSCore.Math.Random@double,double` are included using the path `DSCore.Math.Random`.
- For nodes that are specified in `LayoutSpecs.json` but are not present in `RawTypeData.json` or vice versa, warnings will be displayed on the console.

@Benglin


src/LibraryUtilities.ts
```diff
@@ -175,22 +190,52 @@ export function constructLibraryItem(
...
+                if (isLeafNode && parentNode && (parentNode != result)) {
+                    // If the new node created is a leaf node, append it as a child of 
+                    // the current existing node. This step is necessary for nodes that
+                    // have overloads with the same name but different parameters.
+                    result.appendChild(parentNode);
+                    nodeFound = true;
+
+                    // Reset the value of parentNode, to check through the rest of typeListNodes.
+                    parentNode = inclusive ? null : result;
+                }
```
Benglin: Let's talk through this on Monday and discuss how this can be simplified.

src/LibraryUtilities.ts
```diff
@@ -175,22 +190,52 @@ export function constructLibraryItem(
...
-            // the current resulting node.
-            result.appendChild(parentNode);
+        // If a node that matches the includePath cannot be found
+        if (!nodeFound) { 
+
+            // isLeafNode is set to true if there are no nodes that can match the includePath,
+            // or if the node is a leaf node.
+            if (isLeafNode) {
+                console.warn("Warning: The type '" + includePath + "' is not found in " +
+                    "'RawTypeData.json'. No node of this type is rendered in the library view.");
```
Benglin: `RawTypeData.json` and `LayoutSpecs.json` is just an example JSON file names we happen to use to demo `librarie.js` functionality. In real world when a client consumes `librarie.js`, it does not care about the files at all, the JSON contents will be passed to `librarie.js` in JSON format. This can be simplified as:

```js
console.warn("Matching type not found: " + includePath);
```

src/LibraryUtilities.ts
```diff
@@ -249,7 +294,18 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+
+    let libraryTreeItems: ItemData[] = convertToLibraryTree(typeListNodes, layoutElements);
+
+    // Search for the nodes that are not displayed in library view.
+    _.each(typeListNodes, function (node) {
+        if (!node.processed) {
+            console.warn("Warning: '" + node.contextData + "' is not specified in " +
+                "'LayoutSpecs.json'. The node is not rendered in the library view.");
+        }
+    });
```
Benglin: Same as above this can be simplified as:

```js
console.warn("Item filtered out from library view: " + node.contextData);
```

Benglin:
Thanks @yeexinc, we'll discuss more tomorrow in person, need to talk through some of the changes.

src/LibraryUtilities.ts
```diff
@@ -175,22 +190,52 @@ export function constructLibraryItem(
...
+        let nodeFound: boolean = false;
+        let isLeafNode: boolean = true;
 
         for (let j = 0; j < typeListNodes.length; j++) {
-
             let fullyQualifiedName = typeListNodes[j].fullyQualifiedName;
-            if (!fullyQualifiedName.startsWith(includePath)) {
-                continue; // Not matching, skip to the next type node.
+            let fullyQualifiedNameParts = fullyQualifiedName.split('.');
+            let lastIncludedPart = includeParts[includeParts.length - 1];
```
Benglin:
```js
// Consider the following scenario with each variable value being:
//   "fullyQualifiedNameParts" = [ "DSCore", "Math", "RandomList" ]
//   "includeParts" = [ "DSCore", "Math", "Random" ]
//   "lastIncludedPart" = "Random"
```

src/LibraryUtilities.ts
```diff
@@ -175,22 +190,52 @@ export function constructLibraryItem(
...
+
+            // Check if a matching node is found, and if the fullyQualifiedName contains
+            // the (whole) last word in includePath. 
+            // E.g. This ensures EllipseArc nodes to be excluded in the Ellipse category.
+            if (fullyQualifiedName.startsWith(includePath) && _.contains(fullyQualifiedNameParts, lastIncludedPart)) {
```
Benglin: We can't use `_.contains` here, e.g. the following scenario:

`[ "DSCore", "Math", "Randomizer", "Random" ]` will be included when it should not have been.

yeexinc:
Hi @Benglin , I've updated the codes and added some test cases for this. Thanks!

test/libraryUtilitiesTest.ts
```diff
@@ -314,6 +314,178 @@ describe('constructLibraryItem function', function () {
     expect(result.childItems[0].childItems.length).to.equal(0);
```
Benglin: Thanks so much for adding these test cases! 👍

src/LibraryUtilities.ts
```diff
@@ -203,6 +248,23 @@ export function constructLibraryItem(
...
 
+// E.g. Consider the following scenario with each variable value being:
+// "fullyQualifiedName" = "DSCore.Math.RandomList"
+// "includePath" = DSCore.Math.Random
+// "fullyQualifiedNameParts" = [ "DSCore", "Math", "RandomList" ]
+// "includeParts" = [ "DSCore", "Math", "Random" ]
+// Comparing fullyQualifiedName and includePath with only startsWith() will return true,
+// which is undesired since Random and RandomList are two different nodes.
+// Further checking is required to match the two strings using matchedUpTo().
+function matchedUpTo(fullyQualifiedNameParts: string[], includeParts: string[]): boolean {
```
Benglin: Documentation of an API should not include things that are irrelevant (e.g. why does the reader need to know `startsWith`?). We should simply give scenarios that will make the API returns `true` or `false`, by reading the documentation its usage should become really clear.

Also please use the documentation format that is similar to the following `convertToLibraryTree` API.

src/LibraryUtilities.ts
```diff
@@ -175,22 +190,52 @@ export function constructLibraryItem(
...
-            parentNode = constructNestedLibraryItems(includeParts,
-                typeListNodes[j], inclusive, parentNode, layoutElement.include[i].iconUrl);
+            // If a matching node is found, check if the parts of includePath match those
+            // of fullyQualifiedName.
+            let fullyQualifiedNameParts = fullyQualifiedName.split('.');
+            if (matchedUpTo(fullyQualifiedNameParts, includeParts)) {
+
+                // Check if the includePath represents a leaf node.
+                if (includePath.length < fullyQualifiedName.length)
+                    isLeafNode = false;
```
Benglin: Let's discuss again soon about `isLeafNode` :)

yeexinc:
Hi @Benglin , I modified the logic so that `isLeafNode` is not needed. Please let me know what you think. Thanks!

yeexinc:
This PR only gives warnings in the console if there are left-over items, please refer to #25 for displaying the items in library.

src/LibraryUtilities.ts
```diff
@@ -115,13 +118,25 @@ export function constructNestedLibraryItems(
     for (let i = startIndex; i < fullNameParts.length; i++) {
```
Benglin: Please consider making `for (i)` and `for (j)` into `for (.. of.. )` loop each.

yeexinc:
@Benglin Thanks very much for taking the time to review this! 
The issues will be addressed in the new PR, so I'm closing this one.

PR #14, opened 2017-03-22
Bundle css/react/react-dom all into build

fanwgwg:
- Stylesheet and dependencies are removed from `index.html`, because `LibraryStyles.css`, `react.min.js` and `react-dom.min.js` are now all bundled into `librarie.js`/`librarie.min.js`, depending on build environment
- The font is bundled into`./dist/v0.0.1/resources` folder becasue it is directly referenced in `LibraryStyles.css`

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,11 +1,11 @@
 const webpack = require('webpack');
-let productionBuild = (process.env.production_build == 1);
+let productionBuild = (process.env.NODE_ENV == "production");
 let version = "v0.0.1";
 let plugins = [];
 
 plugins.push(
     new webpack.DefinePlugin({
-        'process.env.NODE_ENV': productionBuild ? JSON.stringify('production') : JSON.stringify('development'),
+        'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
         'global': {}
     })
 );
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -3,6 +3,13 @@ let productionBuild = (process.env.production_build == 1);
 let version = "v0.0.1";
 let plugins = [];
 
+plugins.push(
+    new webpack.DefinePlugin({
+        'process.env.NODE_ENV': productionBuild ? JSON.stringify('production') : JSON.stringify('development'),
+        'global': {}
+    })
+);
+
```
fanwgwg: Yes! For all the dependcies that `imports` or `requires`, it will just bundle them all into the output js file

webpack.config.js
```diff
@@ -3,6 +3,13 @@ let productionBuild = (process.env.production_build == 1);
 let version = "v0.0.1";
```
fanwgwg: I find that `DefinePlugin` is still somehow needed. When we bundle all the reactjs together, it seems that react won't know what is `process`, because `process` is a global variable in Node environment. Since react runs in browser, `process` will not exist there. Therefore, when I already renamed then env variable to `NODE_ENV`, there will be error saying that `process is not defined` when I run `node ./index.js`.

I googled a bit on this, and turned out that many people actually had this issues, and `DefinePlugin` seems to be a workaround for this. Btw, `DefinePlugin` is shipped with `webpack` itself, so this won't introduce extra dependencies. 

Now I've made it look like this 
```javascript
    new webpack.DefinePlugin({
        'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
        'global': {}
    })
```

Benglin: Got it!

PR #15, opened 2017-03-24
add search timeout of 300ms

fanwgwg:
*[DYN-684](https://jira.autodesk.com/browse/DYN-684) Add a timer to search view so that it will only search after user stops typing after 300 ms*

* Search happens immediately when user types something, but change will only be shown on ui after user stops typing for 300ms
* Fixed some failed testcases due to renaming

@Benglin


src/components/SearchView.tsx
```diff
@@ -35,12 +36,22 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
     }
 
     onTextChange(event: any) {
+        clearTimeout(this.timeout);
+
         let text = event.target.value.trim().toLowerCase();
+
+        // start searching only after user stops typing for some time
+        this.timeout = setTimeout(function () {
+            this.updateSearchState(text);
```
Benglin: `updateSearchState` should be done immediately.

src/components/SearchView.tsx
```diff
@@ -35,12 +36,22 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
...
+    }
+
+    updateSearchState(text: string) {
         let hasText = text.length > 0;
 
         this.setState({ searchText: text });
 
-        // Update LibraryContainer of the search
+        // Update LibraryContainer of current search
         this.props.onSearchModeChanged(hasText);
```
Benglin: Only this should go into the time out region.

fanwgwg:
@Benglin changes are there now, search will happen immediately after user types something, but change will only be shown after 300ms. Also, due to renaming in previous PR, there is one testcase failing, I just fixed it here.

src/components/SearchView.tsx
```diff
@@ -86,25 +87,40 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
     }
 
     onTextChange(event: any) {
+        clearTimeout(this.timeout);
```
Benglin: Would you think we need to set this value to an invalid number (e..g `-1`) after clearing the timer? The same question can be asked in constructor too (e.g. `timeout: number = -1`)?

Benglin:
Thanks @fanwgwg for the great work and sorry for the delay! I have left a note, for now I'll merge this in.

src/components/SearchView.tsx
fanwgwg: Just I checked the api doc of `setTimeout` on MDN, I found that in their example, they didn't assign timeout with any value when declaring or clearing it. Api link is here https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

PR #16, opened 2017-03-27
Searchbar component

yeexinc:
*[DYN-515](https://jira.autodesk.com/browse/DYN-515) Implement UI components for search*

This PR adds a SearchBar component into the library view.
captures1

Expanded:
captures2

- When "Display as structured view" or "Display detailed info" is selected, it toggles the `structured` state or `detailed` state of `SearchView` respectively.
- When one of the options in "Filter by category" is selected, it updates the `selectedCategories` state in `SearchView` so that it includes all the categories checked.

@fanwgwg @Benglin


src/components/SearchBar.tsx
```diff
@@ -0,0 +1,91 @@
...
+        }
+        else {
+            categories.push(event.target.name);
+        }
+        this.setState({ selectedCategories: categories})
+        this.props.onCategoriesChange(categories);
+    }
+
+    render() {
+        let options = <div></div>;
```
fanwgwg: Is it better to make `options` equal to `null` rather than `<div></div>`?

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,91 @@
...
+        let searchOptionsBtn = <button id="SearchOptionsBtn" className="ArrowBg" onClick={this.onOptionsButtonClick.bind(this)}></button>;
+        let thisObj = this;
+        var checkboxes = [];
+        _.each(this.props.categories, function (c) {
+            let x = {c};
```
fanwgwg: What's the `x` doing here?

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,91 @@
...
+        }
+        this.setState({ selectedCategories: categories})
+        this.props.onCategoriesChange(categories);
+    }
+
+    render() {
+        let options = <div></div>;
+        let searchOptionsBtn = <button id="SearchOptionsBtn" className="ArrowBg" onClick={this.onOptionsButtonClick.bind(this)}></button>;
+        let thisObj = this;
+        var checkboxes = [];
```
fanwgwg: It's better to decalre the type of a variable as much as possible, so it should be `JSX.Element[]`

yeexinc:
@fanwgwg Thanks very much for the review! I've made the changes.

yeexinc:
Update on the UI:
captures3

riteshchandawar:
wow, looking nice, get it in before our Sprint demo.

Thanks,
Ritesh

yeexinc:
@riteshchandawar - Could you comment on which UI looks better for the search bar?

searchbar2
I personally think this UI goes better with the current library view, but the UI screenshot in my previous comment is closer to the search bar in Dynamo.

Thanks.

Racel:
@yeexinc - Are you going to add the "Select All" and "Only" options? I would like to have those included in the UI as this went through several rounds of design iteration already. Here is the additonal UI in WPF - the 'Only' option appears as a user hovers over each line in the filter dropdown.

image

Also, we may want to consider switching to icons for the layout options...it is a lot of text to read and the term 'Display' is used twice and appears redundant.


Additionally, I would suggest that you get rid of the scrollbar in the filter menu. Users will likely not notice it, and I think it is an additional unnecessary interaction. Here are a few options:
- Similar to the tooltips, show this UI in WPF
- Show this UI OVER the library (users will switch contexts to make their selections already, so it is not a big deal to cover up the library temporarily)
- Expand the UI completely to remove the scrollbar

yeexinc:
@Racel - Thank you for the input. The icons for the layout options have not yet been added, but I've updated it such that the list is fully expanded and the "only" option is added. There is an "All" option which toggles between selecting all items and deselecting all items:
searchbargif

riteshchandawar:
@yeexinc  Maybe you can make that All left align and change it to "Select All", does that make sense @Racel?

Racel:
@riteshchandawar @yeexinc - I think that "All" should be "Select All" and appear visually different than all the other categories...

Maybe put it on the same line as "Filter by category:" and right align it like 'only'

Also, I would try and make "Filter by category" appear differently than the categories. Maybe try making it bold?

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
+import * as React from 'react';
+import * as _ from 'underscore';
+
+export interface SearchBarProps {
+    onTextChange: Function;
+    onStructuredModeChanged: Function;
+    onDetailedModeChanged: Function;
+    onCategoriesChange: Function;
+    categories: string[];
+}
```
Benglin: TASK: turn all `Function` usage here into some function prototypes that outline the parameters/return types.

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+        this.state = { expanded: false, selectedCategories: this.props.categories, structured: false, detailed: false };
+    }
+
+    onTextChange(event: any) {
+        this.props.onTextChange(event);
+    }
+
+    onOptionsButtonClick() {
+        this.setState({ expanded: !this.state.expanded });
+    }
```
Benglin: Can we not call it `Expand` button?

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    onStructuredModeChange(event: any) {
+        this.props.onStructuredModeChanged();
+        this.setState({ structured: !this.state.structured });
+    }
+
+    onDetailedModeChange(event: any) {
+        this.props.onDetailedModeChanged();
+        this.setState({ detailed: !this.state.detailed });
+    }
```
Benglin: Do we not need a parameter for `onStructuredModeChanged` and `onDetailedModeChanged` that carries the current value of `structured` and `detailed` values, respectively?

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+    onAllCheckboxChange(event: any) {
+        if (event.target.checked) {
+            this.setState({ selectedCategories: this.props.categories });
+            this.props.onCategoriesChange(this.props.categories);
+        }
+        else {
+            this.setState({ selectedCategories: [] })
+            this.props.onCategoriesChange([]);
+        }
+    }
```
Benglin: Name change: `onCategoriesChanged` and `onAllCheckboxChanged` (all *changed* should be ending with *ed*).

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+        this.props.onCategoriesChange(categories);
+    }
+
+    allCategoriesSelected() {
+        return (this.state.selectedCategories.length == this.props.categories.length);
+    }
+
+    isSelectedCategory(category: string) {
+        return _.contains(this.state.selectedCategories, category);
+    }
```
Benglin: `isCategorySelected`

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    setSelectedCategory(event: any) {
+        this.setState({ selectedCategories: [event.target.name] })
+    }
```
Benglin: `setSelectedCategories`, it is `selectedCategories` we are updating, it should suggest *more than one* categories.

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    createCheckbox(name: string, checkboxClassName: string, checked: boolean, onChangeFunc: any, displayText?: string): JSX.Element {
```
Benglin: TASK: Put `name`, `checkboxClassName`, `checked`, `onChangeFunc` and `displayText` inside a `class` or an `interface` called `CategoryData`

Benglin: The new class/interface can contain a method that generates the corresponding `JSX.Element`.

Benglin: This way we can get rid of `isSelectedCategory` and a separate `selectedCategories` member.

src/components/SearchView.tsx
```diff
@@ -19,30 +21,50 @@ interface SearchViewProps {
...
 
     constructor(props: SearchViewProps) {
         super(props);
 
-        // set default state
-        this.state = ({
-            searchText: "",
-            displayMode: "list"
-        })
+        this.categories = ["List", "Input", "Math", "Script", "Display", "ImportExport", "String", "Geometry"];
```
Benglin: A `SearchView` component does not even care about these list, it's the caller's job.

src/resources/UI/check-symbol.svg
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
Benglin: No awesome font equivalent?

Benglin: Same for the following ones.

Benglin:
Thanks @yeexinc, I put some `TASK` comments above indicating that we need tasks created for them (also list the task link next to the comment), and some minor tweaks should conclude this review.

src/resources/UI/check-symbol.svg
yeexinc: I think we don't have font-awesome installed here, do we?

src/components/SearchBar.tsx
yeexinc: Task: https://jira.autodesk.com/browse/DYN-840

src/components/SearchBar.tsx
yeexinc: Task link: https://jira.autodesk.com/browse/DYN-841

yeexinc:
Thanks for the review @Benglin , here is the UI update:
searchbarui2
The category options are generated based on the library items available.

Racel:
Thanks for the UI updates @yeexinc! Looks good!

Is there a glow on the 'Filter by category'? Screenshot looks a bit strange..

yeexinc:
The "Filter by category" was bolded, but I guess it wasn't displayed very nicely on the browser.

Here is the updated look with the font "Artifakt Element":
artifakt

src/resources/UI/check-symbol.svg
Benglin: Then the question is, can we install it? No worries for now, can you make one task to introduce *font-awesome* and replace relevant SVG files with it? Thanks!

src/components/SearchView.tsx
```diff
@@ -29,23 +35,49 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
...
     generateStructuredItems(): JSX.Element[] {
+        let structuredItems: JSX.Element[] = [];
+        let categoryItems: ItemData[] = [];
+        
         let index = 0;
-        return this.props.sections.map((item: ItemData) =>
-            <LibraryItem key={index++} libraryContainer={this.props.libraryContainer} data={item} />);
+
+        this.props.sections.forEach(section =>
+            categoryItems = categoryItems.concat(section.childItems)
```
Benglin: Can we not use `categoryItems.push(...)` instead of `categoryItems.concat(...)`? The latter creates a new array altogether, we should use the fastest way.

src/components/SearchView.tsx
```diff
@@ -29,23 +35,49 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
...
+        );
+
+        for (let item of categoryItems) {
```
Benglin: Let's bring `index = 0` to above this line.

Benglin:
Thanks again, @yeexinc, this is good stuff! You can address those minor tweaks later. For now I'm gonna merge this in.

src/resources/UI/check-symbol.svg
yeexinc: Here's the task link: https://jira.autodesk.com/browse/DYN-848

src/components/SearchView.tsx
yeexinc: If `.push(...)` is used, the codes will be:
```
for (let section of this.props.sections) {
    for (let childItem of section.childItems) {
        categoryItems.push(childItem);
    }
}
```
I think `.concat(...)` can give a better performance than the double loop.

PR #17, opened 2017-04-03
Add a search event to allow call to web search component

fanwgwg:
- A search event is created in `LibraryView.tsx`. Upon every search with display mode in `List`, this event will be raised. If it successfully gets JSON contents from the web, it will update the `searchResultListItems` in `SearchView`. When rendering `SearchView`, it will check whether `searchResultListItems` is `null`, if not `null` it will construct the `listItems` from the JSON contents, otherwise it will perform search locally, and generate items based on the display mode.
- Default display mode is now set to be `List`
- Add `SearchResultItem` component for `list` mode
- String in search will now be highlighted
- Gif for the search in `list` mode is shown below

capture

@Benglin


src/EventHandler.ts
```diff
@@ -20,11 +20,13 @@ export class Event {
         this.callbacks.push(callback);
     }
 
-    executeCallback(params?: any | any[]) {
+    executeCallback(params: any[]) {
```
yeexinc: @fanwgwg May I know why is this changed? The function should be able to take in array parameters as it is.

fanwgwg: This is because I tried to allow the function `raiseEvent` to be able to receive multiple parameters, so I changed it to something like what's shown below.
```js
raiseEvent(name: string, ...params: any[]) {}
```
The `...params: any[]` will take in any number of parameters and store them in an array. That's why later in the `executeCallback` function I just make the parameter to be an array.

index.html
```diff
@@ -31,9 +31,21 @@
...
+            let searchText = searchView.getSearchText();
+            console.log(searchText);
+
+            // call search function (from web component) to return json.
+            // If json available, pass the json to search view.
+            // Otherwise pass null to search view, and search view will 
+            // run search locally
+            let searchResultItems = null;
+            searchView.updateSearchResultListItems(searchResultItems);
+        })
```
Benglin: In my opinion, I think `function (searchView)` should be `function (searchText)`, a search algorithm should not need to know about the view in anyway. For this function, it should return the search result to the caller (which is `librarie.js`). So for this case, it will be:

```js
libView.on("searchStarted", function (searchText) {
    return null; // Indicates that library should do its own search
});
```

Benglin: And the event should be named `searchTextUpdated`, and documented in `ReadMe.md` file.

src/LibraryUtilities.ts
```diff
@@ -295,3 +297,24 @@ export function searchItemResursive(items: ItemData[], text: string) {
         search(text, item);
     }
 }
+
+export function getHighlightedText(text: string, highlightedText: string) {
```
Benglin: While you're at it, can you give it a return type?

src/LibraryUtilities.ts
```diff
@@ -295,3 +297,24 @@ export function searchItemResursive(items: ItemData[], text: string) {
...
+        return text;
+    }
+
+    var regex = new RegExp('' + highlightedText + '', 'gi');
+    var segments = text.split(regex);
+    var replacements = text.match(regex);
+    var spans = [];
+    let keyIndex = 0;
+    for (let i = 0; i < segments.length; i++) {
+        spans.push(React.DOM.span({ key: keyIndex++ }, segments[i]));
```
Benglin: You can directly use `i` instead of another new variable `keyIndex`.

src/LibraryUtilities.ts
```diff
@@ -295,3 +297,24 @@ export function searchItemResursive(items: ItemData[], text: string) {
...
+
+export function getHighlightedText(text: string, highlightedText: string) {
+    if (highlightedText.length == 0) {
+        return text;
+    }
+
+    var regex = new RegExp('' + highlightedText + '', 'gi');
+    var segments = text.split(regex);
+    var replacements = text.match(regex);
+    var spans = [];
```
Benglin: Can we use `let` instead of `var` here? Also for `spans` we need its type to be specified (if possible).

src/LibraryView.tsx
```diff
@@ -75,8 +75,6 @@ export class LibraryView {
             libraryView={this}
             loadedTypesJson={this.loadedTypesJson}
             layoutSpecsJson={this.layoutSpecsJson} />, htmlElement);
-
-
```
Benglin: This file can be reverted.

src/LibraryUtilities.ts
```diff
@@ -295,3 +297,24 @@ export function searchItemResursive(items: ItemData[], text: string) {
         search(text, item);
     }
 }
+
+export function getHighlightedText(text: string, highlightedText: string) {
+    if (highlightedText.length == 0) {
+        return text;
+    }
+
+    var regex = new RegExp('' + highlightedText + '', 'gi');
```
Benglin: What does adding an empty character at front/back mean?

src/components/SearchView.tsx
```diff
@@ -1,8 +1,11 @@
 import * as React from "react";
 import { LibraryItem } from "./LibraryItem";
+import { SearchResultItem } from "./SearchResultItem";
 import { LibraryView } from "../LibraryView";
 import { searchItemResursive, setItemStateRecursive, ItemData } from "../LibraryUtilities";
 
+type displayMode = "strucutre" | "list";
```
Benglin: Should be `structure`

src/EventHandler.ts
Benglin: Thanks @yeexinc for helping to review. @fanwgwg, if this requires documentation change please also update `ReadMe.md` and other affected files.

Benglin:
Blown away, @fanwgwg, this is crazy! Thanks! @Racel @kronz, we don't need any more reasons than this to say, "we're going the web way"!

index.html
fanwgwg: If we just pass searchText to this event, and return the search result to libraryView instead of searchView, does it mean that we need to add one more attribute, like `searchResultItems`, to the configuration of libraryView?

src/LibraryUtilities.ts
fanwgwg: Oh at first I added slash in the quotation marks, later I found that it's not needed. I'll remove them.

fanwgwg:
@Benglin Thanks for the review! I've made changes accordingly.

Racel:
@fanwgwg @Benglin - very nice addition.

index.html
Benglin: Why is that we need that extra attribute `searchResultItems`?

Benglin:
Hi @fanwgwg, sorry that you have to resolve the merge conflict now, we can merge it in soon

index.html
fanwgwg: Oh it is already removed in the last commit.

fanwgwg:
@Benglin I've just resolved the conflicts.

PR #18, opened 2017-04-03
Hide image when image load fails

fanwgwg:
For now the broken image will be hidden when its load fails.
We can add some other icon for this special case later.

*[DYN-737](https://jira.autodesk.com/browse/DYN-737) Replace broken image with icons when icon loading fails*

capture

@Benglin


Benglin:
Great stuff, @fanwgwg! Thanks for all the initiatives to fix these little issues! Can you please create a new task for making a replacement icon work? When you do that please add a link here, then we'll merge this in.

fanwgwg:
Yes the link is there now in the comment.

PR #19, opened 2017-04-05
Relace lines with arrows for indentation

fanwgwg:
- Removed prop `indentLevel` from `LibraryItem` component
- Use arrow to indicate indentation, which will change its direction when the item is expanded
- Add `bodyIndentation` for `LibraryItemBodyContainer`, which contains `{clusteredElements}` or `nestedElements`

**Gif that shows the indentation in normal library view**

capture1

### DIFF
--- a/src/components/ClusterView.tsx
+++ b/src/components/ClusterView.tsx
@@ -53,7 +53,7 @@ export class ClusterView extends React.Component<ClusterViewProps, undefined> {
 
         let index = 0;
         return this.props.childItems.map((item: ItemData) => {
-            return (<LibraryItem key={index++} libraryView={this.props.libraryView} data={item} indentLevel={0} />);
+            return (<LibraryItem key={index++} libraryView={this.props.libraryView} data={item} />);
         });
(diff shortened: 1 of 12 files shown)


Benglin:
Thanks @fanwgwg for this fantastic job! I'm super delighted that `indent` value and `for` loop is now gone! I think the reason that is possible is because the outer level item is the container of those inner level items right? So when outer level item gets indented, the inner ones inherit the offset too?

fanwgwg:
@Benglin Yup that's true! Now it is somehow like recursion, so we don't need to do the for loop anymore. Every item will indent its child itself.

Racel:
@Benglin @fanwgwg - Really great work! But, is it possible to make the UI look like the current WPF library which is a mix of both lines and arrows? It is important to maintain the lines in the UI as they signify to users that the list of nodes under a category or subcategory is complete or there is more to scroll down to. Can you please add them back and make it work like the WPF UI? Let me know if we need a separate meeting to discuss.

PR #22, opened 2017-04-12
Update the message returned by click event

fanwgwg:
`contextData` will be passed to event only when a leaf item is being clicked

@sharadkjaiswal


sharadkjaiswal:
Thanks @fanwgwg this looks good.

PR #23, opened 2017-04-19
Simplify the LibraryContainer APIs

Benglin:
### Purpose

This pull request simplifies the API set for `LibraryContainer` (an API set that replaces the existing `LibraryView`). Some notable changes:

- The only way to interact with `LibraryContainer` API now is to provide actual JSON objects rather than the URLs from where these objects are to be downloaded. This is because the `LibraryContainer` makes no assumption from where these JSON objects can be obtained.

- `JsonDownloader` helper class is introduced to assist in downloading JSON objects so the example continues to work as it does today.

- Former `Reactor` object in `LibraryView` has now been moved to `LibraryContainer`, since the former is now deprecated.

- New APIs `CreateLibraryByElementId` and `CreateLibraryContainer` are introduced.

### Declarations

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [ ] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.
 
### Reviewers

@sharadkjaiswal 

### FYIs

@ellensi
@fanwgwg
@yeexinc


sharadkjaiswal:
LGTM

PR #25, opened 2017-04-21
Add "Miscellaneous" section to display left-over items

yeexinc:
[DYN-637](https://jira.autodesk.com/browse/DYN-637) Display left-over items in hierarchical view
[DYN-796](https://jira.autodesk.com/browse/DYN-796) Implement section view in librarie.js

This is a follow-up of #13 
miscgif2

- Nodes that have the same names but different parameters are included by specifying the path once, e.g. `DSCore.Math.Random@int` and `DSCore.Math.Random@double,double` are included using the path `DSCore.Math.Random`.
- Nodes that are specified in `LayoutSpecs.json` but are not present in `RawTypeData.json` or vice versa, warnings will be displayed on the console.
- If there are unprocessed nodes in loadedTypes, a "Miscellaneous" section is added at the bottom of the library view to display them.

In the gif, the unprocessed nodes are:
```
- Count@var[]..[]
- ImportFromCSV@string
- ImportFromCSV@string,bool
- DSCore.Formula.Evaluate@string,string[],var[]
- DSCore.Thread.Pause@var,int
- Analysis.Label.ByPointAndString@Autodesk.DesignScript.Geometry.Point,string
- Analysis.AnalysisExtensions.IsAlmostEqualTo@Autodesk.DesignScript.Geometry.UV,Autodesk.DesignScript.Geometry.UV
```

@Benglin


mjkkirschner:
hi @yeexinc - does the library view show the parameters somewhere? (for all nodes or overloads?)

Racel:
@yeexinc @Benglin - Is there a reason we have "left-over" nodes? For OOTB, we shouldn't have these. 

Also, it is very difficult to determine the hierarchy in the gif. Are "DSCore" and "Analysis" supposed to be under Miscellaneous? And are "Lable" and "AnalysisExtensions" supposed to be under "Analysis"? This is unclear...

Maybe we can do a UI review whenever you are ready.

yeexinc:
@mjkkirschner - Currently when the nodes are clicked on, the library view does a `console.log` to print the full name of the node, which includes the parameters.

@Racel - Left-over nodes are due to some nodes being present in `loadedTypes` (the flat list of nodes) but not specified in `layoutSpecs`. They are not rendered in the library view since the library view doesn't know where the nodes should be placed or how the tree should be structured. As such, the library view attempts to display the left-over nodes in another section using all the information it has. 
In this case, it creates the tree structure using the node's name (by splitting it with `.`) Hence `Analysis.Label.ByPointAndString` goes under Analysis > Label > ByPointAndString.
The nodes in the description are just an example, if they're included in `layoutSpecs`, they won't appear under Miscellaneous.

mjkkirschner:
@yeexinc thats my question :) but I understand it's not related to this PR - can we show the arguments so the user knows that overload they are clicking? - or is that a separate task?

yeexinc:
@mjkkirschner - I agree, the arguments should be shown for overloads as per the existing UI in Dynamo.
I think it would be good if we can have a separate task for this.

Benglin:
Good point, @mjkkirschner, I have create a new task for tracking that here:

https://jira.autodesk.com/browse/DYN-812

Hi @Racel, the *left over nodes* section is to help identify what are the nodes we left out during our development time. Once a node doesn't get included in the `layoutSpecs.json`, it will show up there under `miscellaneous` so developer knows to add it into `layoutSpecs.json` file. This will not get to the end-users.

docs/LayoutSpecs.json
```diff
@@ -534,9 +534,6 @@
               "path": "DSCore.Math.Random"
             },
             {
-              "path": "DSCore.Math.Random"
-            },
-            {
```
Benglin: Didn't we do this previously also?

test/libraryUtilitiesTest.ts
```diff
@@ -314,6 +314,178 @@ describe('constructLibraryItem function', function () {
...
+      }
+    }
+
+    The result should have the following structure:
+    - parent item 1 (result)
+      |- parent item 2
+          |- c
+             |- d
+             |- e
+  */
```
Benglin: Do we have another test that has `inclusive = false`? In which case I expect it to be (without `c`):

  - parent item 1
    - parent item 2
      - d
      - e

test/libraryUtilitiesTest.ts
```diff
@@ -314,6 +314,178 @@ describe('constructLibraryItem function', function () {
     expect(result.childItems[0].childItems.length).to.equal(0);
```
Benglin: Thanks so much for adding the test cases, @yeexinc !

src/components/LibraryContainer.tsx
```diff
@@ -23,13 +23,15 @@ export interface LibraryContainerStates {
 export class LibraryContainer extends React.Component<LibraryContainerProps, LibraryContainerStates> {
 
     generatedLibraryItems: any = null;
+    miscItems: any = null;
```
Benglin: Can both `generatedLibraryItems` and `miscItems` be given specific types here?

src/LibraryUtilities.ts
```diff
@@ -295,7 +359,78 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+    if (fullyQualifiedNameParts.length == 1)
+    {
+        let newNode: ItemData = new ItemData(fullyQualifiedNameParts[0]);
+        newNode.contextData = typeListNode.contextData;
+        newNode.iconUrl = typeListNode.iconUrl;
+        newNode.itemType = typeListNode.memberType;
+        parentNode.appendChild(newNode);
+        return;
+    }
+    let slicedParts = fullyQualifiedNameParts.slice(1, fullyQualifiedNameParts.length);
```
Benglin: Can you add comments about an example of `fullyQualifiedNameParts` and `slicedParts`? That helps understanding code better. Also, can you use just `slice(1)` without the second parameter?

src/LibraryUtilities.ts
```diff
@@ -295,7 +359,78 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+    typeListNode.fullyQualifiedName = slicedParts.join('.');
+
+    // Check through the parent's child items to see if a node of the same name already exists
```
Benglin: Small note here, I found the `for (.. of..)` loop is quite useful in making this easier to read, try it!

Benglin: Again an example value will help describe this way better than readers imagining it.

src/LibraryUtilities.ts
```diff
@@ -295,7 +359,78 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+            buildLibraryItemsFromName(typeListNode, parentNode.childItems[i]);
+            return;
+        }
+    }
+
+    // Otherwise, create a new parent node
+    let newParentNode = new ItemData(fullyQualifiedNameParts[0]);
+    newParentNode.contextData = typeListNode.contextData;
+    newParentNode.iconUrl = typeListNode.iconUrl;
+    newParentNode.itemType = "group";
```
Benglin: Why is it a `group` and not a `none`? We need examples for these codes.

src/LibraryUtilities.ts
```diff
@@ -295,7 +359,78 @@ export function buildLibraryItemsFromLayoutSpecs(loadedTypes: any, layoutSpecs:
...
+    _.each(miscNode.childItems, function(node) {
+        if (node.itemType == "group") node.itemType = "category";
+    })
+
+    libraryTreeItems.push(miscNode);
+    return libraryTreeItems;
+}
+
+/**
+ * Convert an array of typeListNodes to ItemData based on their fullyQualifiedNames.
```
Benglin: If this method converts type list nodes to item data, why is it called `convertToMiscNodes`? We need example data in description (and basically everywhere else) to make the code understandable.

src/LibraryUtilities.ts
```diff
@@ -250,6 +286,34 @@ export function constructLibraryItem(
 }
 
 /**
+ * Compare two arrays of strings and check if the strings in the first array match
+ * those in the second array.
+ * 
+ * @param {string[]} parts
```
Benglin: Can we use names like `firstParts` and `secondParts` to make the method more generic? For `returns` documentation, if you can provide few examples and their corresponding return values? Can we return `-1` if no match is found instead of `undefined`?

src/LibraryUtilities.ts
```diff
@@ -250,6 +286,34 @@ export function constructLibraryItem(
...
+ * @returns
+ * Returns 0 if the two inputs match completely.
+ * Returns how many items 'parts' has more than 'partsToInclude', provided that
+ * 'parts' contains all items in 'partsToInclude'.
+ * Returns undefined if the two inputs do not match.
+ * 
+ */
+function leftoverPart(parts: string[], partsToInclude: string[]): number | undefined {
+    let leftover = parts.length;
+    for (let i = 0; i < partsToInclude.length; i++) {
```
Benglin: Perhaps `for (.. of..)` loop can be better here.

src/LibraryUtilities.ts
Benglin: And the name of method can be `compareParts`?

Benglin:
Hi @yeexinc, almost there! Let's tidy some loose ends, and then I'll have a final pass on it before we merge in. Thanks for all the hard work and patience!

Racel:
Thanks for the comments @Benglin @yeexinc. I understand.

Can we schedule a UX/UI review for look/feel? Let me know when you are ready.

src/LibraryUtilities.ts
yeexinc: There was no particular reason for it to be a `group`, but the node resulted from this is always a parent node. (e.g. if `fullyQualifiedName` = `A.B.C`, this part is recursively executed to create node `A` and `B`).
Since the `iconURL` of these parent nodes are only provided in `layoutSpecs`, I think they should have `group` type so that their icons are not shown.

src/LibraryUtilities.ts
yeexinc: I think the `i` is necessary since the index needs to be used to compare both `parts` and `partsToInclude`.

docs/LayoutSpecs.json
yeexinc: This was in the previous PR which was not merged.

src/LibraryUtilities.ts
yeexinc: Can we keep the names as `parts` and `partsToInclude`, since the two parameters are not interchangeable?
This is a method to replace `matchUpTo`, to check if `parts` contains all strings in `partsToInclude` (and if `parts[0] == partsToInclude[0]`,  `parts[1] == partsToInclude[1]` and so on). I've added some examples into the comments.

yeexinc:
@Benglin - Thanks very much again for the detailed review! I've added some test cases as required and made the changes.

Benglin:
Thanks a lot for bearing with me being nit-picky, this is great work, keep it up! 👍

PR #26, opened 2017-04-21
Update the webpack config file to support modules.

ramramps:
This PR updates the Output.LibraryTarget to "umd".
This is a way for the library to work with all module definitions (and where aren’t modules at all). It will work with commonjs, amd and as global variable.

@Benglin

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,3 +1,4 @@
+"use strict";
 const webpack = require('webpack');
 let productionBuild = (process.env.NODE_ENV == "production");
 let version = "v0.0.1";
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -1,3 +1,4 @@
+"use strict";
```
ramramps: this is required for old webpack. It's a good way to have this. the errors are not super friendly and it never say the package is out of date.

webpack.config.js
```diff
@@ -32,7 +33,7 @@ module.exports = {
         filename: productionBuild ? "librarie.min.js" : "librarie.js",
         path: __dirname + "/dist/" + version + "/",
         publicPath: "./dist/" + version,
-        libraryTarget: "var",
+        libraryTarget: "umd",
```
ramramps: http://tomasalabes.me/blog/_site/web-development/2016/04/30/Webpack-Series-Part-1.html 
Check this site...it is explained very clearly.

Benglin:
Thanks a ton, @ramramps! I appreciate how you dug into this and made it work, save us all the troubles of researching into it. 👍

PR #28, opened 2017-04-24
Show installed packages in add-on section of Library UI

fanwgwg:
[DYN-795 Denote packages in LoadedDataTypes.json](https://jira.autodesk.com/browse/DYN-795)
[DYN-797 Render packages in a separate section of librarie.js](https://jira.autodesk.com/browse/DYN-797)

- Package nodes are denoted in `RawTypeData.json` with `pkg://` being the prefix of `fullyQualifiedName`
- Currently, package nodes are only grouped by their `fullyQualifiedName`. For example, if the `fullyQualifiedName` of a node is `pkg://LunchBox.Geometry.Curves.EndPoints`, it will be grouped as 
```
| LunchBox
| - Geometry
| -- Curves
| --- EndPoints
```
- All the package nodes are shown under `Add-on` section
- All the package nodes are available in search
- Screenshot as below

capture1
capture

### Reviewer
@sharadkjaiswal 

### FYI
@Benglin


src/LibraryUtilities.ts
```diff
@@ -206,14 +164,14 @@ export function constructLibraryItem(
     result.constructFromLayoutElement(layoutElement);
 
     // Traverse through the strings in 'include'
-    for (let i = 0; i < layoutElement.include.length; i++) {
+    for (let include of layoutElement.include) {
```
Benglin: Good to know, thanks!

src/LibraryUtilities.ts
```diff
@@ -271,14 +227,8 @@ export function constructLibraryItem(
...
     layoutElements: LayoutElement[]): ItemData[] {
-    let results: ItemData[] = []; // Resulting tree of library items.
 
-    // Generate the resulting library item tree before merging data types.
-    for (let i = 0; i < layoutElements.length; i++) {
-
-        let layoutElement = layoutElements[i];
-        results.push(constructLibraryItem(typeListNodes, layoutElement));
-    }
+    let results = layoutElements.map(layoutElement => constructLibraryItem(typeListNodes, layoutElement));
```
Benglin: Nice clean-up!

Benglin:
Hi @fanwgwg thanks for the great work! I have a general concern that `librarie.js` now knows a lot about packages (i.e. it even has `pkg://` prefix just for packages). I'm a little worried that the solution isn't scalable and future proof, let's discuss about this tomorrow morning with @sharadkjaiswal and @yeexinc. We may need to find another way.

PR #29, opened 2017-04-26
Raise two mouse events for library items

Randy-Ma:
### Purpose

This is to raise mouse events when the mouse hovers over on library items or leaves them.
- "itemHoveredOn"
- "itemMouseLeave"

BTW, string properties are defined for these event names and an old one - "itemClicked".

### Reviewers
@sharadkjaiswal 

### FYIs
@Benglin

### DIFF
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -128,7 +128,8 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
 
         return (
             <div className={this.getLibraryItemContainerStyle()}>
-                <div className={"LibraryItemHeader"} onClick={this.onLibraryItemClicked.bind(this)} >
+                <div className={"LibraryItemHeader"} onClick={this.onLibraryItemClicked.bind(this)} 
+                        onMouseOver={this.onLibraryItemHoveredOn.bind(this)} onMouseLeave={this.onLibraryItemMouseLeave.bind(this)}>
                     {arrow}
                     {iconElement}
                     <div className={libraryItemTextStyle}>{this.props.data.text}</div>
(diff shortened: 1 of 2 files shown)


src/components/LibraryItem.tsx
```diff
@@ -245,4 +246,18 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
             libraryContainer.raiseEvent("itemClicked", this.props.data.contextData);
         }
     }
+
+    onLibraryItemHoveredOn() {
+        let libraryContainer = this.props.libraryContainer;
+        if (this.props.data.childItems.length == 0) {
+            libraryContainer.raiseEvent("itemHoveredOn", this.props.data.contextData);
```
sharadkjaiswal: I was thinking if we can define string constants on library containers to define all the possible event names so that clients can easily figure out which events are available and also we can get rid of all the string hard coding for event names and any potential errors could be caught at compile time.

Randy-Ma: This looks more like something that should be documented: "If you want to respond to something, then you need to register some event".

sharadkjaiswal: Documenting stuff is a good idea, but that won't help us catch errors at compile time.

Randy-Ma: This may only cause runtime errors if any. But let me try if I can define some common constant strings.

src/entry-point.tsx
```diff
@@ -15,6 +15,10 @@ export function CreateLibraryController() {
 
 export class LibraryController {
 
+    ItemClickedEventName = "itemClicked";
+    ItemHoveredOnEventName = "itemHoveredOn";
```
sharadkjaiswal: I am not choosy about names, but wondering why not `ItemMouseOverEventName`?

Randy-Ma:
@sharadkjaiswal 
Thanks for reviewing the code. As discussed, merging this.

PR #30, opened 2017-04-27
Response to a different mouse event and pass y position as well

Randy-Ma:
### Purpose

This is a folow-up for https://github.com/DynamoDS/librarie.js/pull/29.

Here as the onMouseOver event can be triggered multiple times even it is on the same item, onMouseEnter event will be listened instead to avoid unnecessary computation.

Also the y postion of the item is obtained and passed as well.

### Reviewers
@Benglin @sharadkjaiswal 

### FYIs
@riteshchandawar @monikaprabhu

### DIFF
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -17,6 +17,7 @@
 */
 
 import * as React from "react";
+import * as ReactDOM from "react-dom";
 import { ClusterView } from "./ClusterView";
 import { LibraryContainer } from "./LibraryContainer";
 import { ItemData } from "../LibraryUtilities";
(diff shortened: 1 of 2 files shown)


src/components/LibraryItem.tsx
```diff
@@ -248,18 +251,20 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         }
     }
 
-    onLibraryItemHoveredOn() {
+    onLibraryItemMouseLeave() {
         let libraryContainer = this.props.libraryContainer;
         if (this.props.data.childItems.length == 0) {
-            libraryContainer.raiseEvent(libraryContainer.props.libraryController.ItemHoveredOnEventName,
-                this.props.data.contextData);
+            libraryContainer.raiseEvent(libraryContainer.props.libraryController.ItemMouseLeaveEventName, true);
```
sharadkjaiswal: I think it will make sense to send the context data with MouseLeave event as well so that client won't have to cache and it can figure out which item mouse has left.

src/components/LibraryItem.tsx
```diff
@@ -248,18 +251,20 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
         }
     }
 
-    onLibraryItemMouseLeave() {
+    onLibraryItemMouseEnter() {
         let libraryContainer = this.props.libraryContainer;
         if (this.props.data.childItems.length == 0) {
-            libraryContainer.raiseEvent(libraryContainer.props.libraryController.ItemMouseLeaveEventName, true);
+            var rect = ReactDOM.findDOMNode(this).getBoundingClientRect();
+            let middle = (rect.bottom + rect.top)/2.0;
```
sharadkjaiswal: should we align our tooltip with the top position of the item or the middle? To me, top position makes better sense.

src/components/LibraryItem.tsx
Randy-Ma: There is no cache anywhere even if the context data is not passed.

src/components/LibraryItem.tsx
Randy-Ma: I am also thinking about this and will change it to top.

PR #31, opened 2017-04-27
Add sections to layoutSpec

fanwgwg:
- `sections` are defined in `LayoutSpecs.json`. By default, the first section would be the default section that contains normal library items, and the second section would be the miscellaneous section that contains all the miscellaneous library items. 
- Section header of the default section is hidden.
- All sections are expanded by default. Upon click on section header, it will be expanded/collapsed (except for default section, whose header is hidden).

**Screenshot**
capture
capt1ure

### Reviwer
@Benglin 

### FYI
@sharadkjaiswal @yeexinc


fanwgwg:
@Benglin 
I've updated the code. Also, I've grouped those items without categories in `Miscellaneous` section under `Others` category, so that it doesn't look too messy.

Benglin:
Thanks for the great work, @fanwgwg! Let me take a look at this.

@ramramps we are changing the `LayoutSpecs.json` somewhat, I'll add documentation about that in a bit (which I will need @ellensi help with imagery).

Hi @Racel, some more lights have been shed on the section views we discussed while you were here. Specific design can be tweaked later with your inputs.

docs/LayoutSpecs.json
```diff
@@ -1,1365 +1,1390 @@
 {
```
Benglin: I'd suggest right before I merge your PR, manually add `sections` to the latest `LayoutSpecs.json` so that we won't be having too much conflicts moving forward.

src/LibraryUtilities.ts
```diff
@@ -316,97 +325,105 @@ function compareParts(parts: string[], partsToInclude: string[]): number {
...
  * A tree of hierarchical data type identifiers. This tree is constructed 
  * based entirely on the loaded data types and their fully qualified names.
  * See TypeTreeNode for more information.
  * 
- * @param {LayoutElement[]} layoutElements
- * A tree serving as layout specifications from which library item tree is 
- * constructed. The specifications also contain information of how a given 
- * data type is merged into the resulting library item tree node.
+ * @param section
+ * The section that the converted items will be under
```
Benglin: The type should be `LayoutElement`

src/LibraryUtilities.ts
```diff
@@ -316,97 +325,105 @@ function compareParts(parts: string[], partsToInclude: string[]): number {
...
  *      |- C
  *      |- D
  * 
- * @param unprocessedNodes 
- * The nodes to be converted into itemData 
+ * @param allNodes 
+ * All the typeListNodes
+ * 
+ * @param section
+ * The section that the converted items will be under
```
Benglin: Please help to add the types here, too.

PR #32, opened 2017-04-27
Rename creation to create

fanwgwg:
ItemType `creation` is now renamed to `create`
### Reviewer
@sharadkjaiswal


sharadkjaiswal:
Thanks @fanwgwg, this looks good.

PR #33, opened 2017-04-28
Add dynamo icon for icon that fails to load

fanwgwg:
capture

### Reviwer
@sharadkjaiswal


Benglin:
It would be nicer if we have 32x32px icon (native resolution without scaling). But I'm gonna merge this in for now, thanks @fanwgwg

PR #34, opened 2017-04-28
Add icons for ATF nodes

ellensi:
@gohb


Benglin:
Thanks so much @ellensi!

PR #36, opened 2017-04-28
Display nodes included in other sections

yeexinc:
- Process the nodes included as `path` in layout specification for other sections
- Added some package nodes generated from Dynamo in loadedTypes
sections

@ellensi @Benglin


src/LibraryUtilities.ts
```diff
@@ -371,11 +371,53 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
...
+            if (node.fullyQualifiedName.startsWith(prefix.path)) {
+                let newName = (prefix.path.indexOf("://") == -1) ? node.fullyQualifiedName : node.fullyQualifiedName.split(prefix.path)[1];
+
+                // Copy all data over
+                let newTypeListNode = new TypeListNode(newName);
+                newTypeListNode.fullyQualifiedName = newName;
+                newTypeListNode.contextData = node.contextData;
+                newTypeListNode.iconUrl = node.iconUrl;
+                newTypeListNode.memberType = node.memberType;
+                newTypeListNode.processed = true;
```
Benglin: Why do we need to create a new `TypeListNode` here? As requested before, documentation (comments) on the codes is very important for maintenance purposes. Having to guess what is the purpose here makes things harder for people who ends up reading these.

src/LibraryUtilities.ts
```diff
@@ -371,11 +371,53 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
...
+    })
+
     results.push(convertToMiscSection(typeListNodes, miscSection));
 
     return results;
 }
 
+function convertToOtherSection(typeListNodes: TypeListNode[], section: LayoutElement): ItemData {
+    let sectionData = convertSectionToItemData(section);
+    let prefixes = section.include;
```
Benglin: Can we rename `prefixes` to `includePatterns`? We may want to expand the meaning of `include` in the near future with `path` and probably `prefix` also.

src/LibraryUtilities.ts
```diff
@@ -371,11 +371,53 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
...
+    })
+
+    _.each(nodeToProcess, function (node) {
+        buildLibraryItemsFromName(node, sectionData);
+    })
+
+    // Change the itemType of the outermost parents
+    _.each(sectionData.childItems, function (node) {
+        if (node.itemType === "group") node.itemType = "category";
+    })
```
Benglin: This last loop here exists in `convertToMiscSection` and this method. Can they be moved to the caller instead? For example, do this for each section when everything is all built up (so we don't have to do this in each of these smaller methods).

src/LibraryUtilities.ts
```diff
@@ -371,11 +371,53 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
     let miscSection = sections.find(x => x.text == "Miscellaneous");
 
     results.push(convertToDefaultSection(typeListNodes, defaultSection));
+
+    _.each(sections, function (section) {
+        if (section.text != "default" && section.text != "Miscellaneous") {
+            results.push(convertToOtherSection(typeListNodes, section));
+        }
+    })
```
Benglin: Can you help to consolidate all `default` and `Miscellaneous` strings into the same place as `libraryContainer.props.libraryController.ItemClickedEventName` and use it here? So when we change one place, everywhere else changes. Thanks!

src/LibraryUtilities.ts
Benglin: I am guessing that is because the needs of removing `pkg://` from the FQN. In that case can we do this:

```js
if (node.fullyQualifiedName.startsWith(prefix.path)) {
    let oldName = node.fullyQualifiedName;
    let tempName = node.fullyQualifiedName.split(prefix.path)[1];
    node.fullyQualifiedName = tempName;

    // Insert under section first
    buildLibraryItemsFromName(node, sectionData);

    // Then change back to the old name
    node.fullyQualifiedName = oldName;
}
```

Does that cause any problem?

yeexinc:
@Benglin - Thanks very much for the review! Please let me know what you think of the changes.

yeexinc:
As discussed with @riteshchandawar , here is how the UI look:
capture4
- Revit nodes are displayed in default section if the nodes are available in Dynamo
- The icons for sections are hidden (in case we need them later)
- All left-over nodes have been categorized so the Miscellaneous section is not displayed

src/LibraryUtilities.ts
sharadkjaiswal: I wouldn't advise to change the fullyQualifiedName with the tempName and then change it back. For some reason, if the method buildLibraryItemsFromName throws an exception, then the fullyQualifiedName of the node will not be restored.

sharadkjaiswal:
Thanks @yeexinc, this looks good. I am merging it.

yeexinc:
@sharadkjaiswal - Thank you for the review!

PR #38, opened 2017-05-02
Allow search in keywords

fanwgwg:
- `keywords` will be read from `rawTypeData.json`, and converted to a string array separated by `,`  stored in `ItemData`
- The substring search will go to `keywords` to find matches.
- `contextData` will also be stored in `keywords` for search
- Search string will be highlighted. For example, if the search string is `file.fromPa`, `fromPa` in `fromPath` will be highlighted, as shown in the gif.

capture2

### Reviewer
@sharadkjaiswal 

### FYI
@Benglin

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -14,6 +14,7 @@ export class TypeListNode {
     iconUrl: string = "";
     contextData: any = "";
     memberType: MemberType = "none";
+    keywords: string = "";
     processed: boolean = false;
 
     constructor(data: any) {
@@ -21,6 +22,7 @@ export class TypeListNode {
         this.iconUrl = data.iconUrl;
         this.contextData = data.contextData;
         this.memberType = data.itemType;
+        this.keywords = data.keywords;
     }
 }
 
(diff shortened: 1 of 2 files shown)


riteshchandawar:
Get it in before Sprint Demo, thanks, Ritesh

src/LibraryUtilities.ts
```diff
@@ -494,6 +495,17 @@ function buildLibraryItemsFromName(typeListNode: TypeListNode, parentNode: ItemD
...
 }
 
+// Get keywords from typeListNode and push them into itemData
+export function pushKeywords(itemData: ItemData, typeListNode: TypeListNode) {
+    let keywords = typeListNode.keywords.split(",");
+    keywords.forEach(keyword => {
+        itemData.keywords.push(keyword.toLowerCase().trim())
+    });
+
+    let index = typeListNode.contextData.indexOf("@");
```
sharadkjaiswal: Didn't we change contextData to type `any`? contextData may not be a string.

fanwgwg: Ok then it should be fine if I just use `fullyQualifiedName` here?

fanwgwg:
@sharadkjaiswal I've made changes accordingly.

PR #39, opened 2017-05-02
using query selector to select child elements

ramramps:
Using ```document.querySelector``` to select the child element based on the given parent element. 
This PR solves the problem of having different parent Id with same child Id. This is one of the use case for AVP.

@Benglin


Benglin:
Thanks @ramramps, you rock!

PR #41, opened 2017-05-03
Construct nested library items if partial paths are provided

yeexinc:
- Correctly construct the nested library items if the paths in `include` are partial, e.g. if `path` is `a.b` and the node name is `a.b.c.d`, the node should be correctly nested as `a > b > c > d`
- Added some test cases to check for this as well as for `convertToDefaultSection` and `convertToMiscSection`
- As discussed with @riteshchandawar , the `Analysis` node is removed.

@sharadkjaiswal

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -261,8 +261,22 @@ export function constructLibraryItem(
 
             if (k >= 0) {
                 nodeFound = true;
-                parentNode = constructNestedLibraryItems(includeParts,
-                    typeListNodes[j], inclusive, parentNode, layoutElement.include[i].iconUrl);
+
+                if (k > 1) {
+                    // If the includePath does not represent the immediate parent of the leaf node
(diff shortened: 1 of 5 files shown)


src/LibraryUtilities.ts
```diff
@@ -261,8 +261,22 @@ export function constructLibraryItem(
...
-                    typeListNodes[j], inclusive, parentNode, layoutElement.include[i].iconUrl);
+
+                if (k > 1) {
+                    // If the includePath does not represent the immediate parent of the leaf node
+                    // E.g. includePath = "A.B", fullyQualifiedName = "A.B.C.D"
+                    // The library structure should be constructed such that B contains C, which
+                    // contains the leaf node D.
+
+                    // If the path is not inclusive, start constructing from "C.D" (B will not be created)
+                    // Otherwise, start constructing using "B.C.D"
```
Benglin: Nice documentation, keep it up! 👍

Benglin:
Thanks for the fix and also those well written test cases!

PR #42, opened 2017-05-04
Bring user back to library from search result when category name is clicked

fanwgwg:
[*DYN-517 - Bring user back to category view from search result when category name is clicked*](https://jira.autodesk.com/browse/DYN-517)

- Category name in search result is underlined to indicate that it is clickable
- Upon category name being clicked, text on search bar will be cleared, and the library tree that contains that particular item will be expanded.
- Tooltip will also be shown when mouse is hovering an item in search result now.

capture2

### Reviewer 
@Benglin 

### FYI
@sharadkjaiswal @riteshchandawar


src/LibraryUtilities.ts
```diff
@@ -620,3 +620,17 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
 
     return spans;
 }
+
```
Benglin: Need proper documentation (with example data) for this method.

src/components/LibraryItem.tsx
```diff
@@ -140,7 +140,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         }
 
         return (
-            <div className={this.getLibraryItemContainerStyle()}>
+            <div className={this.getLibraryItemContainerStyle()} >
```
Benglin: This file should be excluded.

src/LibraryUtilities.ts
Benglin: And... of course, test cases.

fanwgwg:
@Benglin I've added some comments and testcases.

src/LibraryUtilities.ts
```diff
@@ -640,3 +640,37 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
...
+    item = allItems.find(item =>
+        item.text == pathToItem[0].text && item.iconUrl == pathToItem[0].iconUrl
+    );
+
+    if (pathToItem.length == 1) {
+        return item ? true : false;
+    } else {
+        pathToItem.shift();
+        item.expanded = true;
+        return findItemByPath(pathToItem, item.childItems);
```
Benglin: Shouldn't this be like:

```js
pathToItem.shift();
let result = findItemByPath(pathToItem, item.childItems);
item.expanded = result; // Expand only if item is found.
return result;
```

src/LibraryUtilities.ts
```diff
@@ -640,3 +640,37 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
...
+ *      |- C
+ *      |- D
+ * 
+ * @param {ItemData} allItems
+ * An array of ItemData, which contains all the items. In the example above,
+ * allItems would be [A]
+ * 
+ * @return {boolean} true if an item is found, false otherwise
+ */
+export function findItemByPath(pathToItem: ItemData[], allItems: ItemData[]): boolean {
```
Benglin: Good documentation, it is from this we can better understand what the function does. The function has a side-effect (item expanded if found), but the name does not suggest that. So I think the name should be `findAndExpandItemByPath`.

src/components/SearchResultItem.tsx
```diff
@@ -19,16 +21,23 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
 
     render() {
         let iconPath = this.props.data.iconUrl;
+        let parentText = this.props.pathToItem[this.props.pathToItem.length - 2].text;
+        let categoryText = this.props.pathToItem.find(item => item.itemType === "category").text;
```
Benglin: Lack of documentation, please improve on this. This section is growing and will soon become incomprehensible if we don't document it properly. What does `pathToItem.length - 2` mean? How is `categoryText` derived? It cannot be understood unless we have proper document with examples here.

src/components/SearchResultItem.tsx
```diff
@@ -41,7 +50,29 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
...
         this.props.libraryContainer.raiseEvent("itemClicked", this.props.data.contextData);
     };
+
+    onLibraryItemMouseLeave() {
+        let libraryContainer = this.props.libraryContainer;
+        if (this.props.data.childItems.length == 0) {
+            libraryContainer.raiseEvent(libraryContainer.props.libraryController.ItemMouseLeaveEventName,
+                { data: this.props.data.contextData });
+        }
+    }
```
Benglin: Very minor thing, I find this more readable:

```js
  let mouseLeave = libraryContainer.props.libraryController.ItemMouseLeaveEventName;
  libraryContainer.raiseEvent(mouseLeave, { data: this.props.data.contextData });
 }
```

What do you think?

fanwgwg:
@Benglin I've made changes addressing your comment.

src/LibraryUtilities.ts
```diff
@@ -644,3 +644,38 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
...
+ * allItems would be [A]
+ * 
+ * @return {boolean} true if an item is found, false otherwise
+ */
+export function findAndExpandItemByPath(pathToItem: ItemData[], allItems: ItemData[]): boolean {
+    let item: ItemData;
+
+    item = allItems.find(item =>
+        item.text == pathToItem[0].text && item.iconUrl == pathToItem[0].iconUrl
+    );
```
Benglin: It's interesting you're comparing between two `iconUrl`, it is highly possible for many items to use the same icons. Is `pathToItem[0].text` not sufficient?

fanwgwg: Since text here is not fullyQualifiedName, but just the title of the item, it won't be sufficient as there are already items with the same title in one group.

fanwgwg: I think it would be better to compare context data, but since contextData might be changed to an object at any time, I'm afraid that this comparison won't hold under such case, because there is not a perfect way to do object comparison in javascript yet.

src/components/SearchResultItem.tsx
```diff
@@ -20,10 +22,18 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
 
     render() {
         let iconPath = this.props.data.iconUrl;
+
+        // Parent of the item is the last but one item in the array pathToItem
```
Benglin:
```js
// The parent of a search result item is the second last entry in 'pathToItem'
```

src/components/SearchResultItem.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { ItemData, getHighlightedText } from "../LibraryUtilities";
 import { LibraryContainer } from "./LibraryContainer";
 
 interface SearchResultItemProps {
     data: ItemData;
     libraryContainer: LibraryContainer;
-    category: string;
     highlightedText: string;
+    pathToItem: ItemData[];
+    onParentTextClicked: Function;
```
Benglin: Please use `interface` to declare a prototype for this method, `Function` is too generic, we want TypeScript to warn us if there's anything that's changed.

src/components/SearchView.tsx
```diff
@@ -3,11 +3,9 @@ import * as _ from "underscore";
 import { LibraryItem } from "./LibraryItem";
 import { SearchResultItem } from "./SearchResultItem";
 import { LibraryContainer } from "./LibraryContainer";
-import { searchItemResursive, setItemStateRecursive, ItemData } from "../LibraryUtilities";
+import { searchItemResursive, setItemStateRecursive, findAndExpandItemByPath, ItemData } from "../LibraryUtilities";
```
Benglin: This is getting longer, can you consolidate that into just one `LibraryUtilities`?

```js
import * as LibraryUtilities from "../LibraryUtilities";

// Later on...
let stuff = LibraryUtilities. findAndExpandItemByPath(...);
```

src/components/SearchView.tsx
```diff
@@ -139,32 +137,40 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
...
         }
 
         this.setState({ searchText: text });
-
-        // Update library container of current search
         this.props.onSearchModeChanged(text.length > 0);
     }
 
+    clearSearch() {
+        let searchInput: any = document.getElementById("SearchInputText");
```
Benglin: I'm not sure if `getElementById` or `ref` is the best way to identify a component, but let's leave this as-is.

src/LibraryUtilities.ts
Benglin: Got it

Benglin:
Thanks @fanwgwg! This looks great! 👍

PR #43, opened 2017-05-05
Collapse siblings when expanding one item

fanwgwg:
[*DYN - 823 Implement expanding of one item collapses the other siblings*](https://jira.autodesk.com/browse/DYN-823)

capture2

### Reviewer 
@Benglin 

### FYI
@riteshchandawar

### DIFF
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -25,7 +25,7 @@ import { ItemData } from "../LibraryUtilities";
 export interface LibraryItemProps {
     libraryContainer: LibraryContainer,
     data: ItemData,
-    onExpand?: Function
+    onItemWillExpand?: Function
 }
 
 export interface LibraryItemState {
@@ -195,7 +195,11 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
                     // types of items except ones of type create/action/query.
                     regularItems.map((item: ItemData) => {
                         return (<LibraryItem
-                            key={index++} libraryContainer={this.props.libraryContainer} data={item} onExpand={this.onExpand.bind(this)} />);
+                            key={index++}
+                            libraryContainer={this.props.libraryContainer}
+                            data={item}
+                            onItemWillExpand={this.onItemWillExpand.bind(this)}
+                        />);
                     })
                 }
             </div>
@@ -252,8 +256,9 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         // Toggle expansion state.
         let currentlyExpanded = this.state.expanded;
         if (this.props.data.childItems.length > 0 && !currentlyExpanded) {
-            this.props.onExpand();
+            this.props.onItemWillExpand();
         }
+        
         this.setState({ expanded: !currentlyExpanded });
 
         let libraryContainer = this.props.libraryContainer;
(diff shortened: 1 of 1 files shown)


src/components/LibraryItem.tsx
```diff
@@ -258,6 +263,14 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         }
     }
 
+    // Unexpand all other slibling items when one item is being expanded
```
Benglin: Should be *collapse all other siblings when an item is being expanded*

Benglin: Wait, I think this does something entirely different. Let's rephrase it this way:

```js
// Collapse all child items when one of the child items is expanded
onSingleChildItemExpanded() {
  for (...) { }
  this.setState({ expanded: true }); // Make the current item (parent) expanded.
}
```

src/components/LibraryItem.tsx
```diff
@@ -249,6 +251,9 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     onLibraryItemClicked() {
         // Toggle expansion state.
         let currentlyExpanded = this.state.expanded;
+        if (this.props.data.childItems.length > 0 && !currentlyExpanded) {
+            this.props.onExpand();
+        }
```
Benglin: The notification is sent before expansion actually take place, so let's rename this from `onExpand` to `onItemWillExpand` (just to borrow from React.js term `componentWillReceiveProps` and `componentWillUnmount`, etc).

src/components/LibraryItem.tsx
Benglin: Wait, it should be named `onSingleChildItemWillExpand`, and not `onSingleChildItemExpanded`.

Benglin:
Thanks for this, @fanwgwg! 👍 minor name changes and we'll go with it.

fanwgwg:
@Benglin I've made the changes accordingly! Thanks for the review!

src/components/LibraryItem.tsx
```diff
@@ -193,7 +194,12 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
                     // 'getNestedElements' method is meant to render all other 
                     // types of items except ones of type create/action/query.
                     regularItems.map((item: ItemData) => {
-                        return (<LibraryItem key={index++} libraryContainer={this.props.libraryContainer} data={item} />);
+                        return (<LibraryItem
+                            key={index++}
+                            libraryContainer={this.props.libraryContainer}
+                            data={item}
+                            onItemWillExpand={this.onItemWillExpand.bind(this)}
+                        />);
```
Benglin: Lining up this way is much easier to read! 👍

src/components/LibraryItem.tsx
```diff
@@ -258,6 +268,14 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         }
     }
 
+    // Collapse all child items when one of the child items is expanded
+    onItemWillExpand() {
```
Benglin: Can we rename as I originally recommended? This is at the parent level and we need to make that clear.

fanwgwg:
@Benglin  Ok I've just renamed the function

mjkkirschner:
curious about this - is this an option that can be set at construction time or the only behavior?

Benglin:
Hi @mjkkirschner, do you mean `onItemWillExpand` behaviour or the single-expansion behaviour? Feel free to ping me on Slack, I can chat :)

PR #47, opened 2017-05-11
Implement function prototypes and CategoryData class

yeexinc:
[DYN-840](https://jira.autodesk.com/browse/DYN-840) Turn all Function usage in SearchBar into function prototypes
[DYN-841](https://jira.autodesk.com/browse/DYN-841) Put checkbox attributes in a class or interface called categoryData

- The `Function` in `SearchBar` are replaced with function prototypes which outline the parameters and return types.
- Implement a new class called `CategoryData` which contains a method to create a `checkbox` HTML element.
- When the checkbox is clicked, the `checked` attribute is updated accordingly and the state `selectedCategories` of `SearchBar` is set so that the UI will be updated to reflect the checked values.
- Address some comments in #16


src/components/SearchBar.tsx
```diff
@@ -16,11 +16,32 @@ export interface SearchBarState {
     detailed: boolean;
 }
 
+interface SearchDisplayModeChangedFunc {
+    (value: boolean): void;
+}
```
Benglin: Thanks @yeexinc! One thing though, from looking at this function prototype I am not sure what `value === true` means. Does it mean `structured` or `list` display? This is the reason why `onStructuredModeChanged` and `onDetailedModeChanged` event handlers cannot share the same `SearchDisplayModeChangedFunc`. We need different ones for each of them, and also more meaningful names than `value: boolean` (if it was `structured: boolean`, it tells me that `true` means structured view).

src/components/SearchBar.tsx
```diff
@@ -16,11 +16,32 @@ export interface SearchBarState {
...
+
+interface SearchCategoriesChangedFunc {
+    (categories: string[]): void;
+}
+
+interface SearchTextChangedFunc {
+    (event: any): void;
+}
+
```
Benglin: By right function prototypes should come before `export interface SearchBarProps`.

yeexinc:
Hi @Benglin , I've made the changes. Thanks!

Benglin:
Thanks @yeexinc! LGTM

PR #48, opened 2017-05-11
Library scroll

ellensi:
### Purpose

*[DYN-819](https://jira.autodesk.com/browse/DYN-819) Scroll for the library should be restricted to Library Items only.*

The search bar should not be scrollable together with the library contents!

### Before & After screenshots

image

### Reviewers

@gohb 

### FYIs

@riteshchandawar 

image


Benglin:
Infinite scrolling...

Infinite scrolling...

PR #50, opened 2017-05-12
Display parameters for overloads in search results

yeexinc:
Display parameters for overloads in search results (for list view):
searchparams

@Benglin @fanwgwg


Benglin:
Thanks @yeexinc, LGTM!

PR #51, opened 2017-05-12
Implement search result display modes (compact vs. detailed)

fanwgwg:
[*DYN-519 Implement search result display modes (compact vs. detailed)*](https://jira.autodesk.com/browse/DYN-519)

Node description will be shown under title when `Display detailed info` is selected.

capture2

### Reviewer
@Benglin


src/components/SearchResultItem.tsx
```diff
@@ -22,12 +23,23 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
         let highLightedItemText = getHighlightedText(this.props.data.text, this.props.highlightedText, true);
         let highLightedCategoryText = getHighlightedText(this.props.category, this.props.highlightedText, false);
         let ItemTypeIconPath = "src/resources/icons/library-" + this.props.data.itemType + ".svg";
+        let ItemDescription: JSX.Element = null;
+
+        if (this.props.detailed) {
+            let description = "No description available";
```
Benglin: Is this the current behaviour for Dynamo? I'm thinking `ItemDescription` can remain `null` if there's no description (to simplify the DOM tree), does that make sense? Anyway, both `ItemDescription` and `ItemTypeIconPath` should be camel case.

fanwgwg: Currently, dynamo shows "no description available" if there is no description. Should we follow the current dynamo or just don't show description when there is no description?
capture

Benglin: Okay let's then follow the existing Dynamo behaviour, thanks!

Benglin:
Thanks for the great work, @fanwgwg 👍

PR #53, opened 2017-05-12
Change font format from OTF to WOFF

ellensi:
Because WOFF is designed for web and somehow looks better (probably due to proper hinting).


Racel:
FYI @QilongTang...are we using WOFF on playground?

QilongTang:
@Racel We are using tff

Benglin:
Hi @ellensi, if you can give a more elaborated screen comparison that would make this selection easier. @Racel @QilongTang, can we have a sneak peak into the playground? @ramramps mentioned this to me a while back and I have been curious ever since (pardon me if it is broadcasted on Slack, I am not too up-to-date there).

ellensi:
OTF on the left, WOFF on the right. Note that the **bold** types (blue highlights) are noticeably uneven in OTF, but other differences in non-bold texts are cosmetics only.

image

image

image

image

Note: These screenshots are taken using Chrome on Windows.

Benglin:
Since we are staying with Artifakt font, just a format change (and also the look/feel is visibly better), I'm merging this in for now. Thanks folks!

Racel:
@Benglin - Sending you a link on slack about the playground...

PR #56, opened 2017-05-16
Disable filter button when not needed

fanwgwg:
[*DYN-854 - Filter option should be disabled when there is no active search*](https://jira.autodesk.com/browse/DYN-854)
- Disable filter button when search is not activated
- Disable `Display detailed mode` when in structure mode
- Collapse filter options menu when text is cleared

capture2

### Reviewer
@Benglin 

### FYI
@riteshchandawar


src/resources/LibraryStyles.css
```diff
@@ -199,56 +199,58 @@ input {
...
+    height: 30px;
+    width: 30px;
+    border: none;
+    outline: none;
     background-color: #414141;
 }
 
-.SearchBar #SearchOptionsBtn:hover {
-    background-color: rgba(255, 255, 255, 0.1);
-    cursor: pointer;
```
Benglin: I have a thought about making this consistent when we still can, that is, remove element identifier unless there's a need to identify them by their IDs. Can we have a task to standardise all to use `className` instead?

src/components/SearchBar.tsx
```diff
@@ -57,18 +57,30 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
     }
 
     onTextChanged(event: any) {
         let text = event.target.value.toLowerCase().replace(/ /g,'');
+
+        // collapse filter options menu when text is cleared
+        if (text.length == 0) {
+            this.setState({ expanded: false });
+        }
+
```
Benglin: Can this `setState` be merged with the one right below?

src/components/SearchBar.tsx
```diff
@@ -57,18 +57,30 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
     clearInput() {
         let searchInput: any = document.getElementById("SearchInputText");
         searchInput.value = '';
-        this.setState({ hasText: false });
+        this.setState({
+            hasText: false,
+            expanded: false // collapse filter options menu when text is cleared
+        });
```
Benglin: I'm thinking out loud... but it seems like we can just keep `hasText` and remove `expanded`, if we don't allow expansion when there is no search text. Can this be optimized/streamlined?

fanwgwg: Hmm...when there is no search text, `expanded` is always false. However, when there is text on the search bar, `expanded` can be true or false based on user click.

Benglin: Ah got it.

src/components/SearchBar.tsx
```diff
@@ -57,18 +57,34 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
+        // collapse filter options menu when text is cleared
+        if (text.length == 0) {
+            this.setState({
+                expanded: false,
+                hasText: false
+            });
+        } else {
+            this.setState({ hasText: true });
+        }
+
```
Benglin: How does this look? :)

```js
let expanded = text.length == 0 ? false : this.state.expanded;
this.setState({ expanded: expanded, hasText: text.length > 0 });
```

src/resources/LibraryStyles.css
Benglin: Please make a task and provide this PR link in that Jira task, and then add that task list here.

fanwgwg: JIRA tasks is here: https://jira.autodesk.com/browse/DYN-865

Benglin:
Thanks so much, @fanwgwg ! 👍

PR #58, opened 2017-05-17
Suppress console warnings during tests

yeexinc:
Suppress console warnings so that `console.warn()` will not be displayed during tests:
warns


test/libraryUtilitiesTest.ts
```diff
@@ -256,8 +256,14 @@ describe('constructLibraryItem function', function () {
   beforeEach(function () {
     layoutElements = [];
     typeListNodes = [];
+    console.warn = function() {};
```
Benglin: `// Suppress logging to console in during test runs` would be nice.

Benglin:
Thanks @yeexinc 👍

PR #59, opened 2017-05-17
Remove whitespace from search

fanwgwg:
e.g. Search for `By Coordinates` can return the result that contains `bycoordinates`

### Reviewer
@sharadkjaiswal 

### FYI
@riteshchandawar

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -573,9 +573,9 @@ export function buildLibraryItemsFromName(typeListNode: TypeListNode, parentNode
 export function pushKeywords(itemData: ItemData, typeListNode: TypeListNode) {
     let keywords = typeListNode.keywords.split(",");
     keywords.forEach(keyword => {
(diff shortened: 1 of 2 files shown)


src/LibraryUtilities.ts
```diff
@@ -573,9 +573,9 @@ export function buildLibraryItemsFromName(typeListNode: TypeListNode, parentNode
 export function pushKeywords(itemData: ItemData, typeListNode: TypeListNode) {
     let keywords = typeListNode.keywords.split(",");
     keywords.forEach(keyword => {
-        itemData.keywords.push(keyword.toLowerCase().trim())
+        itemData.keywords.push(keyword.toLowerCase().replace(/ /g, ''));
```
sharadkjaiswal: should we use `replace(/s/g, '')` to search all whitespace characters?

sharadkjaiswal: on second thought it looks ok because keywords won't have other kinds of whitespace characters

PR #55, opened 2017-05-15
Remove SearchView component

yeexinc:
[DYN-852](https://jira.autodesk.com/browse/DYN-852)

- Remove `SearchView` component and move the relevant states (`structured`, `detailed`, `selectedCategories` etc. to `LibraryContainer`).
- Move the `SearchBar` component to `LibraryContainer` so that it exists at the same level as the library items
- Implement a `Searcher` component that contains the utility functions for searching

PR #60, opened 2017-05-17
Make categories available based on search text

Randy-Ma:
### Purpose

This is to make the filtering categories in the expanded search bar dynamic. Only categories containing items as a result of searching will appear.

### Reviewers
@Benglin

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -17,6 +17,10 @@ interface SearchTextChangedFunc {
     (event: any): void;
 }
 
+interface SearchBarExpandedFunc {
+    (event: any): void;
+}
+
 export interface SearchBarProps {
     onTextChanged: SearchTextChangedFunc;
     onStructuredModeChanged: StructuredModeChangedFunc;
@@ -35,7 +39,7 @@ export interface SearchBarState {
 
 export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
 
-    categoryData: CategoryData[] = []
+    categoryData: CategoryData[] = [];
 
     constructor(props: SearchBarProps) {
         super(props);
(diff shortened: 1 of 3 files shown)


src/components/SearchBar.tsx
```diff
@@ -51,7 +51,27 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
             let data = new CategoryData(c, "CategoryCheckbox", true, this.onCategoriesChanged.bind(this));
             data.onOnlyButtonClicked = this.onOnlyButtonClicked.bind(this);
             this.categoryData.push(data);
-        }.bind(this))
+        }.bind(this));
+    }
+
+    //To receive new prop with updated categories
```
Benglin: Does this comment make sense?

```js
// Whenever search text changes in 'SearchView' component, it triggers a call to 
// 'updateSearchView' method after a predetermined delay, which in turn updates
// the properties of 'SearchBar', resulting in this method getting called. This 
// method will attempt to preserve the check states of current categories on to 
// the new categories.
```

src/components/SearchBar.tsx
```diff
@@ -51,7 +51,27 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
...
+        _.each(newProps.categories, function (c: string) {
+            let oldCategory = oldCategoryData.find((x) => {
+                return x.name === c;
+            });
+            let data = new CategoryData(c, "CategoryCheckbox", true, this.onCategoriesChanged.bind(this));
+            data.onOnlyButtonClicked = this.onOnlyButtonClicked.bind(this);
+            if (oldCategory) {
+                data.checked = oldCategory.checked;
+            }
+            this.categoryData.push(data);
```
Benglin: I'd encourage to place `let oldCategory = ...` right before `if (oldCategory) ...` because now they look disconnected.

src/components/SearchView.tsx
```diff
@@ -32,6 +32,10 @@ export class SearchView extends React.Component<SearchViewProps, SearchViewState
 
     searchResultListItems: any;
 
+    //To hold the categories which will be displayed in the search bar.
+    //This will include both the selected and the unselected items.
+    displayedCategories: string[];
```
Benglin: Some suggestions:

```js
    // This list represents the actual categories beind displayed on the search bar 
    // at the time search is performed. It will be a subset of 'props.categories', 
    // that are most relevant to the current search results.
```

Benglin:
Thanks a lot for working on this, @Randy-Ma it looks great! Only thing is, if you can hold off until #55 is merged that'll make things easier to merge.

PR #61, opened 2017-05-18
New APIs introduced for librarie.js

Benglin:
### Purpose

This PR updated the APIs of `librarie.js` to allow for *loaded types* and *layout specifications* JSON objects to be specified at a later time. In another word, `LibraryContainer` can now be created without *loaded types* and *layout specs* JSON objects.

#### Here are the changes in high level:

- `updateSections` utility function and its corresponding test cases are added:

  updateSections function
  - should throw exceptions 0
  - should throw exceptions 1
  - should throw exceptions 2
  - should throw exceptions 3
  - should throw exceptions 4
  - should insert a new element when there isn't an existing one
  - should update existing contents with new contents
  - should update existing contents including child elements

- `LibraryContainer` class now exposes three new methods:

  - `setLoadedTypesJson`
  - `setLayoutSpecsJson`
  - `refreshLibraryView`

- The above new methods cannot be invoked directly through `LibraryContainer`, but should instead be done through `LibraryController` object (which as the same set of new methods)

### Declarations

- [x] The code base is in a better state after this PR
- [x] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [ ] The level of testing this PR includes is appropriate
- [ ] User facing strings, if any, are extracted into `*.resx` files
- [x] All tests pass using the self-service CI.
- [ ] Snapshot of UI changes, if any.

### Reviewers

@sharadkjaiswal 
@fanwgwg 
@yeexinc 

### FYIs

@ramramps

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -504,7 +504,23 @@ function updateElement(oldElement: LayoutElement, newElement: LayoutElement, app
                 });
             }
         }
+    }
 
+    // Done handling 'include' property, proceed to deal with 'childElements'
+    for (let childElement of newElement.childElements) {
+
+        let nameOfChildToUpdate = childElement.text;
+        let childToUpdate = oldElement.childElements.find(function (c) {
+            return c.text === nameOfChildToUpdate;
+        });
+
+        if (!childToUpdate) {
+            // If no existing child is found, insert the new element directly.
+            oldElement.childElements.push(childElement);
+        } else {
+            // If an existing child is found, update it recursively.
+            updateElement(childToUpdate, childElement, append);
+        }
     }
 }
 
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -464,6 +464,9 @@ export function convertSectionToItemData(section: LayoutElement): ItemData {
     return sectionData;
 }
 
+export function updateSections(oldElement: LayoutElement, newElement: LayoutElement, append: boolean): void {
+}
+
 /**
  * Convert an array of typeListNodes to ItemData which are not processed based on their fullyQualifiedNames, 
  * by splitting the name with '.', and put them under a specific section.
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -464,7 +464,44 @@ export function convertSectionToItemData(section: LayoutElement): ItemData {
     return sectionData;
 }
 
-export function updateSections(oldElement: LayoutElement, newElement: LayoutElement, append: boolean): void {
+function updateElement(oldElement: LayoutElement, newElement: LayoutElement, append: boolean): void {
+}
+
+// See 'updateElement' method above for details.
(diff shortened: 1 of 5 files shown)


src/components/LibraryContainer.tsx
```diff
@@ -25,20 +23,85 @@ export interface LibraryContainerStates {
...
+            this.props.defaultSectionString, this.props.miscSectionString);
 
         // Obtain the categories from each section to be added into the filtering options for search
         for (let section of this.generatedSections) {
             for (let childItem of section.childItems)
                 this.searchCategories.push(childItem.text);
         }
+
+        // Just to force a refresh of UI.
+        this.setState({ inSearchMode: this.state.inSearchMode });
```
sharadkjaiswal: will this call trigger UI render even if the state parameter value is not changed?

src/entry-point.tsx
```diff
@@ -13,6 +13,18 @@ export function CreateLibraryController() {
     return new LibraryController();
 }
 
+interface SetLoadedTypesJsonFunc {
+    (loadedTypesJson: any): void;
```
sharadkjaiswal: How about the additional parameter for `append`?

src/entry-point.tsx
```diff
@@ -41,26 +58,46 @@ export class LibraryController {
...
+    createLibraryContainer() {
         return (<LibraryContainer
             libraryController={this}
-            layoutSpecsJson={layoutSpecsJson}
-            loadedTypesJson={loadedTypesJson}
             defaultSectionString={this.DefaultSectionName}
             miscSectionString={this.MiscSectionName} />);
     }
+
+    setLoadedTypesJson(loadedTypesJson: any): void {
```
sharadkjaiswal: We do need a parameter to append. Although, from Dynamo most of the time we would call it with `append: false`.

src/LibraryUtilities.ts
```diff
@@ -464,6 +464,94 @@ export function convertSectionToItemData(section: LayoutElement): ItemData {
     return sectionData;
 }
 
+function updateElement(oldElement: LayoutElement, newElement: LayoutElement): void {
```
sharadkjaiswal: Not sue where is this method being used?

src/components/LibraryContainer.tsx
Benglin: Yes it definitely will (without this the view wasn't updated). I think `React.js` does this because state object may sometimes be object references, even when the reference does not change, its internal values may. So `setState` must trigger a render.

src/entry-point.tsx
Benglin: My glaring oversight, sorry, working on it now!

src/LibraryUtilities.ts
Benglin: It's called from `updateSections` right below it ;)

Benglin:
Thanks for helping with the review, @sharadkjaiswal. Problems fixed.

PR #68, opened 2017-05-23
Reset item expansion when refreshing library view

fanwgwg:
- Sections other than default section are collapsed by default.
- Section `expanded` attribute will be set when being constructed. 

### Reviewer 
@sharadkjaiswal 

### FYI
@Benglin


sharadkjaiswal:
Thanks @fanwgwg this looks good. I am merging it.

PR #70, opened 2017-05-23
Keep item expansion state when refreshing library view

fanwgwg:
When `refreshLibraryView` is called, it will check current expanded items, and apply expansion to new items.

### Reviewer
@sharadkjaiswal @Benglin


src/LibraryUtilities.ts
```diff
@@ -788,7 +788,35 @@ export function findAndExpandItemByPath(pathToItem: ItemData[], allItems: ItemDa
...
+ * 
+ * - A
+ *   |- B
+ *      |- C
+ *      |- D
+ *   |- E
+ *   |- F
+ * 
+ */
+export function getPathToExpandedItemFromRootItem(rootItem: ItemData, itemPath: ItemData[] = []): ItemData[] {
```
Benglin: `getPathToExpandedItem` should be good enough

Benglin: Do we really need to return `ItemData[]` or will `string[]` work as well? I think we need `ItemData[]` for its completeness, but just want to confirm.

Benglin: I think `itemPath` is passed by reference, isn't it? If that's the case we don't really have to return anything from this method (the `itemPath` can be updated in-place).

src/components/LibraryContainer.tsx
```diff
@@ -112,11 +112,23 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
             this.props.defaultSectionString, this.props.miscSectionString);
 
+        // Restore item expansion
+        if (this.generatedSections) {
+            for (let previousSection of this.generatedSections) {
+                if (previousSection.expanded) {
+                    let pathToExpandedItem = LibraryUtilities.getPathToExpandedItemFromRootItem(previousSection);
+                    let newSection = newSections.find(section => section.text == previousSection.text);
+                    LibraryUtilities.findAndExpandItemByPath(pathToExpandedItem, [newSection]);
+                }
```
Benglin: Should we `break` out of this `for` loop once an expanded item is found?

src/components/LibraryContainer.tsx
```diff
@@ -112,11 +112,23 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
     refreshLibraryView(): void {
-
-        this.generatedSections = LibraryUtilities.buildLibrarySectionsFromLayoutSpecs(
+        let newSections = LibraryUtilities.buildLibrarySectionsFromLayoutSpecs(
             this.loadedTypesJson, this.layoutSpecsJson,
             this.props.defaultSectionString, this.props.miscSectionString);
 
+        // Restore item expansion
+        if (this.generatedSections) {
+            for (let previousSection of this.generatedSections) {
```
Benglin: `previousSection` should be renamed to `oldSection` (since you have `newSection`) and then also *previous* sounds like another section from a previous iteration of the loop.

src/components/LibraryItem.tsx
```diff
@@ -269,8 +269,9 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         if (this.props.data.childItems.length > 0 && !currentlyExpanded && this.props.onItemWillExpand) {
             this.props.onItemWillExpand();
         }
-
-        this.setState({ expanded: !currentlyExpanded });
+        
+        this.props.data.expanded = !currentlyExpanded;
+        this.setState({ expanded: this.props.data.expanded });
```
Benglin: Can we add a comment for what line `273` tries to do?

src/LibraryUtilities.ts
```diff
@@ -17,6 +17,7 @@ export class TypeListNode {
     keywords: string = "";
     parameters: string = "";
     description: string = "";
+    isNew: boolean = false;
```
Benglin: Wait, what? No no no no no no no no no no... let's talk.

fanwgwg: Oh the latest commit has some of my code just for testing purpose...I'll push another one later after I resolved some conflicts.

mjkkirschner:
@ramramps @Racel .... this PR seems useful for the addon closing task... unfortunately it is incomplete.
I have another solution, but it changes an old behavior so I consider it slightly risky. I also noted that the addons category is only collapsed on the first creation of a custom node instance as that customNode workspace is only created then.

Might be worth considering finishing this PR instead....

mjkkirschner:
@ramramps @Racel -
I am still looking if there is another fix possible on the dynamo side - what is happening is that when you first click the custom node the entry is added and then removed. - This forces the library to get updated.

The entry is removed because there is a mismatch between the full category name from the customNode workspace - like `xxx.yyy.Query` and the fullCatoryname after processing like `xxx.yyy`. - which is what we actually set the searchElement category name to - we store the Query in a `Group`property and instead of directly with the category. This causes a mismatch and we do a remove and add instead of simply updating the entry.

ramramps:
@mjkkirschner the old searchviewmodel handles it correctly. I am not sure whether this PR should be completed, or add a fix in Dynamo. May be, fix in Dynamo will be easy.

Racel:
@mjkkirschner @ramramps if we are not going to use this approach. Can we close this PR?

QilongTang:
@mjkkirschner Completely obsoleted approach?

PR #71, opened 2017-05-24
Add library header with filter and compact/detail buttons

yeexinc:
[DYN-868 There should be a header for Library and it should have Filter and Compact/Detail buttons.](https://jira.autodesk.com/browse/DYN-868) 

Add a library header with the filter, structured and detailed view options:
**Improved version:**
searchheader3

**Initial implementation:**
searchheader

The buttons are enabled only when user is doing a search.

**Reviewer**
@Benglin


yeexinc:
Update on the UI:
searchheader3

yeexinc:
Updated colour for disabled buttons:
header4

riteshchandawar:
@yeexinc Buttons looks good now.

src/Searcher.tsx
```diff
@@ -133,8 +133,10 @@ export class Searcher {
     }
 
     clearSearch() {
-        let searchInput: any = document.getElementById("SearchInputText");
-        searchInput.value = "";
-        this.clearSearchFunc(searchInput.value);
+        let searchInput: any = document.getElementsByClassName("SearchInputText");
```
Benglin: Let's use `ref` instead, see [this document](https://facebook.github.io/react/docs/refs-and-the-dom.html) for details. Maybe name it as `this.searchInputField`.

Benglin: Do the same for all the other such instances below.

src/components/SearchBar.tsx
```diff
@@ -192,13 +215,18 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
 
         return (
             <div className="SearchBar">
+                <div className="LibraryHeader">
+                    Library
+                    <div>
+                        |{filterBtn}|{structuredBtn}|{detailedBtn}
```
Benglin: Instead of `{filterBtn}` here, you can do `createFilterButton()`, and extract the complexities out of `render` method.

riteshchandawar:
Hello @yeexinc We need to add tooltip for Filter buttons.

You can take a reference of below image to understand it better.
image

yeexinc:
@riteshchandawar @Benglin - Thanks for the review, I've made the changes as discussed.

Here is the update on the dropdown with a slightly bigger font size and padding:
header5

Benglin:
Thanks so much for all your work, @yeexinc! With you leaving Autodesk I hope this won't be your last PR to `librarie.js` repo. Best of luck to your studies, we will send defects along your way to keep your life interesting. Thanks again for all your great contribution! 👍

PR #72, opened 2017-05-24
Implement overlay expandable tooltip

fanwgwg:
[*DY-699 Explore overlay tooltip within HTML*](https://jira.autodesk.com/browse/DYN-699)

- By default, the expandable tooltip is disabled. It can be enabled in the filter options menu.
- When the expandable tooltip is enabled, each leaf item can show its tooltip by clicking on three dots at the right end of the item
- Expandable tooltip is available in both normal library view and search result view
- Tooltip will get data from Dynamo side by raising event `itemToolTipExpand`

capture2

### Reviewer
@Benglin

### FYI
@riteshchandawar


Racel:
@Benglin @riteshchandawar - This feature seems redundant and adds extra UI affordances that are not ideal and I suspect will not be utilized by customers often. I would suggest that we not merge this for now.

Benglin:
Hi @fanwgwg, this PR is too large. Please break it down into multiple PRs (e.g. supporting classes can go in first, and then the other parts). Also, I will review when we make a decision on this:

The original intent was to have an option to turn these tooltip on/off, but not in a way that user can change it. That is, the code that hosts the `librarie.js` will determine whether or not the tooltip is enabled.

fanwgwg:
Close to break this down to some smaller PRs

PR #73, opened 2017-05-24
Add one more property for a event name and update readme document and index.html

Randy-Ma:
### Purpose

This is to add a property in the LibraryController class for "searchTextUpdated" like other event names. The new property is called SearchTextUpdatedEventName.

Meanwhile, the readme document is updated and also index.html is updated.

### Reviewers
@sharadkjaiswal

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ libController.on("someEventName", function(data) {
 });
 ```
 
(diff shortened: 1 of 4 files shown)


sharadkjaiswal:
LGTM

PR #74, opened 2017-05-24
Update font and icon colours

yeexinc:
Update font and icon colours to light grey `#aaa`:
fontcolour1

Expanded items have `#eee` as their font colour:
fontcolour2

@riteshchandawar


riteshchandawar:
Wow, you have to change that in so many places? Anyway, it is looking good now. Thanks for accommodating last min request.

PR #76, opened 2017-05-25
Add showExpandableTooltip state and ToolTip component

fanwgwg:
- A new state `showExpandableToolTip` is added to `LibraryContainer`, and it is set to `false` by default.
- `LibraryItem` and `SearchReusltItem` will also have the props `showExpandableToolTip` inherited from `LibraryContainer`
- A `ToolTip` component is added, which will be rendered by `LibraryItem` and `SearchResultItem` (but not in this PR yet)
- A new state `toolTipExpanded` is added to `LibraryItem`

### Reviewer
@Benglin


Benglin:
Hi @fanwgwg since this is a UI enhancement, let's put an image for it.

src/Searcher.tsx
```diff
@@ -6,18 +6,11 @@ import { LibraryContainer } from "./components/LibraryContainer";
...
 
-type displayMode = "structure" | "list";
-
-interface SearchModeChangedFunc {
-    (inSearchMode: boolean): void;
-}
-
 interface ClearSearchFunc {
     (searchText: string): void;
 }
```
Benglin: Not related to your change but the fact that `ClearSearchFunc` takes a `string` seems a little weird to me.

src/Searcher.tsx
```diff
@@ -42,7 +38,7 @@ export class Searcher {
         this.displayedCategories = categories;
     }
 
-    generateStructuredItems(): JSX.Element[] {
+    generateStructuredItems(showExpandableToolTip: boolean): JSX.Element[] {
```
Benglin: I'm thinking of not calling it a *tool tip* because no matter who looks at it, they will have the conventional floating tooltip window in mind. If we can call it *ItemDescription*, then this parameter can be named `displayItemDescription`.

src/Searcher.tsx
```diff
@@ -6,18 +6,11 @@ import { LibraryContainer } from "./components/LibraryContainer";
 import * as LibraryUtilities from "./LibraryUtilities";
 import { SearchBar } from "./components/SearchBar";
 
-type displayMode = "structure" | "list";
-
-interface SearchModeChangedFunc {
-    (inSearchMode: boolean): void;
-}
-
```
Benglin: The PR description does not tell why this is removed, can you add that reason to the description? Can't seem to recall why is this done...

src/components/LibraryItem.tsx
```diff
@@ -20,26 +20,28 @@ import * as React from "react";
 import * as ReactDOM from "react-dom";
 import { ClusterView } from "./ClusterView";
 import { LibraryContainer } from "./LibraryContainer";
-import { ItemData, sortItemsByText } from "../LibraryUtilities";
+import * as LibraryUtilities from "../LibraryUtilities";
```
Benglin: Nice, thanks!

src/components/LibraryItem.tsx
```diff
@@ -20,26 +20,28 @@ import * as React from "react";
...
-    data: ItemData,
+    data: LibraryUtilities.ItemData,
+    showExpandableToolTip: boolean,
     onItemWillExpand?: Function
 }
 
 export interface LibraryItemState {
-    expanded: boolean
+    itemExpanded: boolean,
+    toolTipExpanded: boolean
```
Benglin: If we end up renaming *item tooltip* to *item description*, then please restore `expanded` name.

Benglin: And name `toolTipExpanded` as `descriptionShown`

src/Searcher.tsx
fanwgwg: I am not sure why this is added for `Searcher`, as it is not used at all...so I just removed it

fanwgwg:
@Benglin I've renamed it to `ItemSummary`. Since this PR is only about the supporting classes, there are no real UI changes here. In the next PR, I'll add the UI part.

src/Searcher.tsx
Benglin: Thanks for cleaning up! 👍

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+    libraryContainer: LibraryContainer,
+    data: ItemData;
+    showDescription: boolean;
+}
+
+interface ItemSummaryStates {
+    hasItemSummaryData: boolean;
+}
+
+export class ItemSummary extends React.Component<ItemSummaryProps, ItemSummaryStates> {
```
Benglin: Good naming choice 👍

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+    itemSummaryData: any;
+
+    constructor(props: ItemSummaryProps) {
+        super(props);
+        this.state = ({ hasItemSummaryData: false });
+        this.itemSummaryData = null;
+        this.onReceiveDataFromDynamo = this.onReceiveDataFromDynamo.bind(this);
```
Benglin: Oops, the `librarie.js` has nothing to do with Dynamo (which is just one of its consumers), so a better name would be `onDataReceived` or `handleReceivedData`.

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+            let itemSummaryExpandEvent = libraryContainer.props.libraryController.ItemSummaryExpandEventName;
+            libraryContainer.raiseEvent(
+                itemSummaryExpandEvent,
+                { dataReceiver: this.onReceiveDataFromDynamo, data: this.props.data.contextData }
+            );
+        }
+    }
+
+    // Data received from Dynamo side will be a tuple with Item1 being input parameters, 
+    // Item2 being output parameters, Item3 being description text.
```
Benglin: Please remove all references of *Dynamo* in `librarie.js`

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+    hasItemSummaryData: boolean;
+}
+
+export class ItemSummary extends React.Component<ItemSummaryProps, ItemSummaryStates> {
+    itemSummaryData: any;
+
+    constructor(props: ItemSummaryProps) {
+        super(props);
+        this.state = ({ hasItemSummaryData: false });
+        this.itemSummaryData = null;
```
Benglin: A very minor thing, since we are already in the class `ItemSummary`, the names of data members can be simplified. For example, `hasSummaryData` and `this.summaryData`.

src/entry-point.tsx
```diff
@@ -30,6 +30,7 @@ export class LibraryController {
     ItemClickedEventName = "itemClicked";
     ItemMouseEnterEventName = "itemMouseEnter";
     ItemMouseLeaveEventName = "itemMouseLeave";
+    ItemSummaryExpandEventName = "iteSummaryExpand";
```
Benglin: Okay you need to add this to `readme.md` too

Benglin: And there's a typo

src/components/LibraryItem.tsx
```diff
@@ -296,7 +303,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         for (let item of this.props.data.childItems) {
             item.expanded = false;
         }
-        this.setState({ expanded: true }); // Make the current item (parent) expanded.
+        this.setState({ itemSummaryExpanded: true }); // Make the current item (parent) expanded.
```
Benglin: I don't think you meant to change this one.

src/components/LibraryItem.tsx
```diff
@@ -89,7 +92,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
             return null;
         }
 
-        let iconElement = null;
+        let iconElement: JSX.Element = null;
```
Benglin: 👍

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    // Raise event to get data from Dynamo side if there is no data yet.
+    onLibraryItemSummaryExpand() {
+        if (!this.state.hasItemSummaryData) {
+            let libraryContainer = this.props.libraryContainer;
+            let itemSummaryExpandEvent = libraryContainer.props.libraryController.ItemSummaryExpandEventName;
+            libraryContainer.raiseEvent(
+                itemSummaryExpandEvent,
+                { dataReceiver: this.onReceiveDataFromDynamo, data: this.props.data.contextData }
```
Benglin: `dataReceiver` sounds like an object, what about `onDataReceivedHandler`?

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+        let descriptionText = this.props.data.description;
+        let input: JSX.Element[] = [];
+        let output: JSX.Element = null;
+        let description: JSX.Element = null;
+        let icon: JSX.Element = null;
+
+        if (this.state.hasItemSummaryData) {
+            let inputParameters = this.itemSummaryData.Item1;
+            let outputParameters = this.itemSummaryData.Item2;
+            let descriptionTextFromDynamo = this.itemSummaryData.Item3;
```
Benglin: Having names like `Item1` etc. is a side-effect of serialization process of Dynamo (one of several clients), we should not allow that side-effect to leak all the way to `librarie.js`, this needs to be made something more meaningful. They need to be updated to `inputParameters`, `outputParameters` and `descriptionText`, and clients like Dynamo will have to conform to this (to be documented) format.

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+            for (let inputParameter of inputParameters) {
+                let inputParameterName = inputParameter.Item1
+                if (inputParameterName.length > 0) {
+                    inputParameterName.concat(": ");
+                }
+
+                input.push(<div className={"IOName"}>{inputParameterName}: {inputParameter.Item2}</div>);
+            }
+
+            output = <div className={"IOName"}>{outputParameters}</div>;
```
Benglin: If there is a chance please update `IOName` to something more meaningful.

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    <div className={"Output"}>OUTPUT</div>
+                    {output}
+                </div>
+                <img className={"Icon"} src={this.props.data.iconUrl} onError={this.onImageLoadFail} />;
+            </div>
+        );
+    }
+
+    onImageLoadFail(event: any) {
+        event.target.src = require("../resources/icons/Dynamo.svg");
```
Benglin: While removing references to `Dynamo`, this needs to be renamed to `default-icon.svg`.

fanwgwg:
@Benglin I've updated the code addressing your comments, thanks a lot for the review!

src/components/SearchResultItem.tsx
```diff
@@ -14,14 +14,18 @@ interface SearchResultItemProps {
     pathToItem: LibraryUtilities.ItemData[];
     onParentTextClicked: ParentTextClickedFunc;
     detailed: boolean;
+    showItemSummary: boolean;
 }
 
-interface SearchResultItemStates { }
+interface SearchResultItemStates {
+    itemSumamryExpanded: boolean
```
ellensi: Wrong spelling here :speak_no_evil:

fanwgwg: Hahha okay I'll fix this!

README.md
```diff
@@ -213,12 +213,24 @@ libController.on("itemMouseLeave", function (searchText) {
...
 
 ```js
-libController.on(libController.ItemMouseLeaveEventName, function(contextData) {
+libController.on(libController.ItemMouseLeaveEventName, function(arg) {
     console.log("Data: " + arg.data);
     console.log("Rect(top, left, bottom, right): " + arg.rect.top + "," + arg.rect.left + "," + arg.rect.bottom + "," + arg.rect.right);
 })
 ```
 
+### Event 'itemSummaryExpand'
```
Benglin: Can we call it `itemSummaryExpanded`? It aligns with all other events (except probably *Enter* and *Leave*, which are somewhat okay because of historical reasons, but we can change those too next time).

README.md
```diff
@@ -213,12 +213,24 @@ libController.on("itemMouseLeave", function (searchText) {
...
+
+This event is raised when used clicks on the expand icon displayed on the right of a leaf library item or search result item. This event should return data for showing summary, including `InputParameters`, `OutputParameters` and `Description`.
```
Benglin: Typo *when user clicks*

README.md
```diff
@@ -228,12 +228,24 @@ libController.on("itemMouseLeave", function (searchText) {
...
     console.log("Data: " + arg.data);
     console.log("Rect(top, left, bottom, right): " + arg.rect.top + "," + arg.rect.left + "," + arg.rect.bottom + "," + arg.rect.right);
 })
 ```
 
+### Event 'itemSummaryExpanded'
+
+This event is raised when user clicks on the expand icon displayed on the right of a leaf library item or search result item. This event should return data for showing summary, including `InputParameters`, `OutputParameters` and `Description`.
+
+- `arg`: This contains a callback function `onDataReceivedHandler` and `data` of the item.
```
Benglin: Okay after reading the code I understood how this `onDataReceivedHandler` is used. There are two issues:

1. The name is misleading, we need to let client knows it is the way for it to send in the data, so `setDataCallback` (or something similar).
2. The `ReadMe.md` is meant for documenting the event handler's usage, we need a good example here. Something like the following:

```js
libController.on(libController.ItemSummaryExpandedEventName, function (arg) {
  var data = arg.contextData;
  var finalData = {
    "InputParameters": getInputParamsForItem(data),
    "OutputParameters": getOutputParamsForItem(data)
    "Description": getItemDescription(data)
  }

  // Send data back to librarie.js
  arg.setDataCallback(finalData);
});
```

3. `arg.data` should be named `arg.contextData` so that it is less ambiguous.
4. And I do think we need `inputParameters` rather than `InputParameters` as a JSON standard.

README.md
```diff
@@ -218,8 +218,8 @@ libController.on(libController.ItemMouseEnterEventName, function(arg) {
 
 This event is raised when the mouse leaves the range of one library item.
 
-```js
-libController.on("itemMouseLeave", function (searchText) {
+ ```js
+libController.on("itemMouseLeave", function (arg) {
     console.log("Data: " + arg.data);
     console.log("Rect(top, left, bottom, right): " + arg.rect.top + "," + arg.rect.left + "," + arg.rect.bottom + "," + arg.rect.right);
 });
```
Benglin: Can we wrap these long lines so that we don't have horizontal scroll bars on read me

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    // Raise event to get data if there is no data yet.
+    onLibraryItemSummaryExpand() {
+        if (!this.state.hasSummaryData) {
+            let libraryContainer = this.props.libraryContainer;
+            let itemSummaryExpandEvent = libraryContainer.props.libraryController.ItemSummaryExpandEventName;
+            libraryContainer.raiseEvent(
+                itemSummaryExpandEvent,
+                { onDataReceivedHandler: this.onDataReceived, data: this.props.data.contextData }
```
Benglin: Just `onDataReceivedHandler` need to be renamed to `setDataCallback`.

Benglin: I'm not sure if `onDataReceived` can be named better as `setItemSummary`?

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    onImageLoadFail(event: any) {
+        event.target.src = require("../resources/icons/default-icon.svg");
+    }
+
+    // Raise event to get data if there is no data yet.
+    onLibraryItemSummaryExpand() {
+        if (!this.state.hasSummaryData) {
+            let libraryContainer = this.props.libraryContainer;
+            let itemSummaryExpandEvent = libraryContainer.props.libraryController.ItemSummaryExpandEventName;
```
Benglin: Can you make a task (and add it here) to fix this:

It feels weird to have `libraryController` define all the event names (e.g. `ItemSummaryExpandEventName`), and have `raiseEvent` method implemented in `libraryContainer`. Can we not have event names and event method both define in the same class (i.e. the most suitable  class which all other classes have access to, maybe `libraryController`)?

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    constructor(props: ItemSummaryProps) {
+        super(props);
+        this.state = ({ hasSummaryData: false });
+        this.summaryData = null;
+        this.onDataReceived = this.onDataReceived.bind(this);
+    }
+
+    render() {
+        this.onLibraryItemSummaryExpand();
```
Benglin: `onLibraryItemSummaryExpand` should be renamed to `fetchMissingItemSummary` so that it does not sound like it will raise an event for external components to do other things.

Benglin: I guess the convention of `onXxyyzz` typically makes other think of an *event handler* method.

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,101 @@
...
+            output = <div className={"IOParameter"}>{outputParameters}</div>;
+
+            if (descriptionTextReceived.length > 0) {
+                descriptionText = descriptionTextReceived;
+            }
+        }
+
+        if (!descriptionText) {
+            descriptionText = "No description available";
+        }
```
Benglin: I believe for times when `this.state.hasSummaryData == false`, we should show *Fetching summary...*? This will be cleared anyway when the summary finally arrives.

src/components/ItemSummary.tsx
```diff
@@ -0,0 +1,101 @@
+import * as React from 'react';
+import { ItemData } from "../LibraryUtilities";
+import { LibraryContainer } from "./LibraryContainer";
+
+interface ItemSummaryProps {
+    libraryContainer: LibraryContainer,
+    data: ItemData;
+    showDescription: boolean;
```
Benglin:
```js
// 'ItemSummary' can optionally display the description text. In regular library 
// view the description text is always shown as part of 'ItemSummary' when it is 
// expanded. In search result view however, 'ItemSummary' does not display its 
// description text as each search result item already displays the same description
// text (in detailed view), in which case 'showDescription' property is set to false. 
```

fanwgwg:
@Benglin I've updated the readme and renamed some of the functions.

fanwgwg:
@Benglin I've created a JIRA task at https://jira.autodesk.com/browse/DYN-881 to put all the event names and raiseEvent method into one class.

Benglin:
Thanks again @fanwgwg and sorry for the delay!

PR #78, opened 2017-05-29
UI bug fix and enhancements

fanwgwg:
- [DYN-824](https://jira.autodesk.com/browse/DYN-874) Pressing Esc key will clear the search text now
- [DYN-871](https://jira.autodesk.com/browse/DYN-871) Icons for Library items won't get shrink now.
- The icon on section header won't be out of place anymore when directing back to normal library view from category text.
- The section header won't split the page anymore when scrolling up and down

### Reviewer
@Benglin 

### FYI
@sharadkjaiswal @riteshchandawar


src/components/SearchBar.tsx
```diff
@@ -246,7 +255,12 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
                 <div className="SearchInput">
                     <div>
                         <i className="fa fa-search SearchBarIcon"></i>
-                        <input className="SearchInputText" type="input" placeholder="Search..." onChange={this.onTextChanged.bind(this)} ref={(field) => { this.searchInputField = field; this.props.setSearchInputField(field) }}></input>
+                        <input
+                            className="SearchInputText"
+                            type="input" placeholder="Search..."
+                            onChange={this.onTextChanged.bind(this)}
+                            ref={(field) => { this.searchInputField = field; this.props.setSearchInputField(field) }}>
+                        </input>
```
Benglin: Hi @fanwgwg can you make a new task to remove `setSearchInputField` from `props`? Unless I'm wrong but I can't imagine why would external components need to have access to the text box (as opposed to dealing with it through a set of APIs exposed from `SearchBar` component). Please add the task and put the link here (and make that task reference to this particular PR comment).

src/components/LibraryItem.tsx
```diff
@@ -89,19 +89,27 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
         // If it is a section, display the icon (if given) and provide a click interaction.
-        if (this.props.data.itemType === "section") {
-            iconElement = this.props.data.iconUrl ? (<img className={"LibraryItemIcon"} src={this.props.data.iconUrl} onError={this.onImageLoadFail}
-                onClick={this.onSectionIconClicked.bind(this)} />) : null;
+        if (this.props.data.itemType === "section" && this.props.data.iconUrl) {
+            iconElement = <img
+                className={"LibraryItemIcon"}
+                src={this.props.data.iconUrl}
+                onError={this.onImageLoadFail}
+                onClick={this.onSectionIconClicked.bind(this)}
```
Benglin: If you can move `onSectionIconClicked.bind` to where `onImageLoadFail.bind` was done that'll make this neater 😄

Benglin:
Add one task, and we'll let this in. Thanks @fanwgwg

src/components/SearchBar.tsx
fanwgwg: I've just created a new task for it: https://jira.autodesk.com/browse/DYN-878

fanwgwg:
@Benglin I've made changed accordingly and added a new task link under your comment.

fanwgwg:
@ellensi Please help merge this PR.

src/resources/LibraryStyles.css
```diff
@@ -397,7 +398,6 @@ input {
 .SearchResultItemContainer .ItemIcon {
     padding: 2px 8px;
     width: 40px;
-    height: 40px;
```
ellensi: Hi @fanwgwg,

Seems like the problem with distorted icons here is caused by the fact that `.ItemInfo` on the right side of the icon space is set to match the width of its parent element (`width: 100%`), hence pushing (and compressing) the icon to the left. Could you please remove the `width: 100%` of `.ItemInfo` instead and leave the `.ItemIcon` with its native dimensions at 32 × 32px? Thanks!

fanwgwg: Yup I realised this. However, the width of `.ItemInfo` will be changed when the description text is shown, which will change the width of `.ItemIcon`, and that's why I set the width of `.ItemInfo` to be 100% so that its width stays unchanged. To keep the native dimension of the icon, I tried that set the width of `.ItemInfo` to be 90% works, and it doesn't affect the dimension of the icon. Is this fine?

ellensi: I see, I haven't seen the description text yet! Could you clarify why `.ItemInfo` width needs to be set as a relative percentage? The problem with relative sizing with an absolute-sized icon will arise when the user tries to resize the window/library view.

fanwgwg: Oh yup it doesn't work on resizing. Another workaround is that we set the `min-width` of `.ItemIcon` to be 40px, so that it won't be pushed to left by `.ItemInfo`, and the percentage width of `.ItemInfo` can also be removed.

ellensi: Yeah, this "workaround" looks better! And can we change the search result icon size to 32px instead of 40px? I guess this size will go better since we decided to make everything in the LibraryUI look smaller in order to match the rest of the Dynamo UI.

fanwgwg: I've just updated the size and removed the percentage width.

ellensi:
Thanks @fanwgwg! Merging this in, as Ben gave the green light.

PR #80, opened 2017-05-30
Add keyboard navigation to search result view

fanwgwg:
[*Add keyboard navigation to search result view*](https://jira.autodesk.com/browse/DYN-875)
- Allow `Up/Down` arrow to navigate around search result items
- Allow `Enter` key to create currently selected node
- Clicking on an item will make this item selected

capture2

### Reviewer
@Benglin 

### FYI
@riteshchandawar


fanwgwg:
@Randy-Ma Can you help review this PR? Thanks!

src/components/LibraryContainer.tsx
```diff
@@ -69,6 +71,29 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
 
+    componentWillMount() {
+        window.addEventListener("keydown", this.handleKeyDown.bind(this));
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener("keydown", this.handleKeyDown.bind(this));
+    }
+
+    handleKeyDown(event: any) {
```
Randy-Ma: Can we handle the event in one place? By any chance, if the event handler in SearchResultItem is handled before this one in LibraryContainer, there will be logic error.

fanwgwg: I've removed the keyboard event in `LibraryContainer`. The first `SearchResultItem` will update the `selectionIndex` in `LibraryContainer` on arrow up/down keyboar event now.

src/components/LibraryContainer.tsx
```diff
@@ -139,7 +141,48 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
+    }
+
+    // Set the max selection index(number of items in the search result).
+    setMaxSelectionIndex(max: number) {
+        this.maxSelectionIndex = max;
+    }
+
     onSearchModeChanged(inSearchMode: boolean) {
+        // Reset selectionIndex when quitting from search mode
+        if (this.state.inSearchMode && !inSearchMode) {
```
Randy-Ma: So long as the parameter is false, we need to update the two index variables, right?

src/components/SearchResultItem.tsx
```diff
@@ -8,23 +8,94 @@ interface ParentTextClickedFunc {
...
+                    this.onItemClicked();
+                }
+            default:
+                break;
+        }
+    }
+
+    // Update the current selectionIndex in libraryContainer
+    updateSelectionIndex(selectNextItem: boolean) {
+        if (this.props.index == 0) {
```
Randy-Ma: Why only updating the selection index when the index of the props is 0? Not for other SearchResultItem items?

fanwgwg: This is to update the selectionIndex of libraryContainer. It makes more sense for LibraryContainer to keep track of the selectionIndex, rather than every SearchResultItem has one attribute for this index. Every time when user presses arrow up/down, all the SearchResultItems would respond to the event. However, libraryContainer only needs to be called once, that's why I only make the first item set the index in libraryContainer.

src/components/SearchResultItem.tsx
```diff
@@ -8,23 +8,94 @@ interface ParentTextClickedFunc {
...
+                currentItem.scrollIntoView();
+            }
+
+            if (currentRect.bottom > containerRect.bottom) {
+                currentItem.scrollIntoView(false);
+            }
+        }
+    }
+
+    handleKeyDown(event: any) {
```
Randy-Ma: Hmmmm, if there are 1000 SearchResultItem items, the event will be handled 1000 times?

src/components/LibraryContainer.tsx
fanwgwg: This function is called even if user just keeps typing whitespaces on the search bar. Under this case, inSearchMode is always set to false, but the index doesn't need to be updated.

src/components/SearchResultItem.tsx
Randy-Ma: I see, I get it, it means the selection index will be only updated once.

src/components/SearchResultItem.tsx
```diff
@@ -8,23 +8,94 @@ interface ParentTextClickedFunc {
...
+
+    // Update the current selectionIndex in libraryContainer
+    updateSelectionIndex(selectNextItem: boolean) {
+        if (this.props.index == 0) {
+            this.props.libraryContainer.updateSelectionIndex(selectNextItem);
+        }
+        this.updateSelectionState();
+    }
+
+    // Select this item if selectioIndex matches the index of this item, and this item is not
```
Randy-Ma: Typo, selectionIndex

src/components/SearchResultItem.tsx
fanwgwg: Since every SearchResultItem is aware of its own state, this would have better performance than having libraryContainer in charge of the selection state of every item. For libraryContainer to do so, it has to rerender the whole 1000 items to refresh the selection state. However, if each SearchResultItem checks against selectionIndex to see whether itself is selected, only the two items that have state changed will be rerendered, other items would only respond to the event but not the rendering.

src/components/LibraryContainer.tsx
```diff
@@ -36,6 +36,8 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
     searchCategories: string[] = [];
 
     timeout: number;
+    selectionIndex: number = -1;
```
Randy-Ma: There is not a selected item by default. The expected behavior should be that when you are searching, and then press Enter, a node will be created from the 1st search result.

src/Searcher.tsx
```diff
@@ -117,6 +118,12 @@ export class Searcher {
             return 0;
         });
 
+        for (let i = 0; i < leafItems.length; i++) {
+            leafItems[i].props.index = i;
+        }
+
+        this.libraryContainer.setMaxSelectionIndex(leafItems.length - 1);
```
Randy-Ma: Do we need to reset selectionIndex as well? Otherwise, when search text changes, the default one will not be the 1st any more.

src/components/LibraryContainer.tsx
fanwgwg: So do you mean that the first searchResultItem should be selected by default?

src/components/SearchResultItem.tsx
Randy-Ma: It won't rerender the whole 1000 items as only items that have been changed will be rerendered. 

Also if it happens that the event handler of the item to be selected is called first, as the selected index is not updated yet, it will never be udpated as selected.

src/components/LibraryContainer.tsx
Randy-Ma: In that case, I am wondering why onSearchModeChanged is called. It should not be called at all, right?

src/components/SearchResultItem.tsx
fanwgwg: If we let LibraryContainer handle the event, it means it should have a state change when selection changed. Since all the items are dynamically created in the render function of LibraryContainer, it will call the function `generateListItems` to generate all the items, so all the items will be generated no matter whether their state are changed or not.

src/components/LibraryContainer.tsx
Randy-Ma: Yes.

But as this is related to UI behavior, let's check what @Racel says.

src/components/SearchResultItem.tsx
fanwgwg: I think maybe the sequence that the event is called is the sequence that the event is registered. Since the items are generated by order, in my hands-on with it, the selection index is always updated as expected...

Randy-Ma: I get what you mean. This is tricky. I am not quite sure when a new list of items are generated, whether they will all be rerendered or rerendered only based on the actual changes in this case. But have you realized the potential logic error.

Randy-Ma:
@Racel 

There is one comment about UI behavior which need your input.

For the search resultant items, should we make the 1st one as the default one so that it will be selected and when the Enter key is pressed, its related itemClicked event will be triggered?

src/components/LibraryContainer.tsx
fanwgwg: Yup I will prevent it from being called, then here we only need to check against the parameter.

src/components/SearchResultItem.tsx
fanwgwg: Actually I tried that way in my first implementation, and I found that all items are being rerendered, which made everything slow, then I changed it to the current implementation.

Racel:
@Randy-Ma - Yes, please make the first one selected so that when the Enter key is pressed that it will be the one that shows up on the screen. Will using the up and down keys when the library is focused move the selection?

fanwgwg:
@Racel Yes, up/down keys will move the selection.

fanwgwg:
@Randy-Ma I've made changes to allow only `LibraryContainer` handle the key event. `LibraryContainer` hold a ref to all the `SearchResultItems`, and it will set their selection state based on its `selectionIndex`.

src/components/SearchResultItem.tsx
```diff
@@ -34,13 +41,13 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
...
     }
 
-    componentWillReceiveProps(nextProps: SearchResultItemProps) {
-        if (nextProps.libraryContainer.selectionIndex == this.props.index) {
-            this.setState({ selected: true });
-        } else if (this.state.selected && nextProps.libraryContainer.selectionIndex != this.props.index) {
-            this.setState({ selected: false });
-        }
-    }
+    // componentWillReceiveProps(nextProps: SearchResultItemProps) {
```
Randy-Ma: If this part is not used, why still keep it?

fanwgwg: I've just removed these few lines.

Randy-Ma:
@fanwgwg Cool! Thanks for the hard work!

PR #81, opened 2017-05-30
Make it possible to use an external search function

Randy-Ma:
### Purpose

This pull request is to make it possible to use external search service or function to search library items based on a given string.

The implementation here defines a new member SearchLibraryItemsFunc in LibraryController. It can be set and upon set, will be used to search.

When SearchLibraryItemsFunc is set and successfully called, it will call the callback function with the result in the format as specified in https://github.com/DynamoDS/librarie.js/blob/master/docs/loaded-data-types.md.

As the resultant items are not in the same priority, a new field priority is added to relevant data types.

### Reviewers
@sharadkjaiswal 

### FYIs
@Benglin


Racel:
@Randy-Ma How will this affect the in-canvas search? Are we changing the fundamental search algorithm? FYI @mjkkirschner

Randy-Ma:
@Racel If we do not set the exposed SearchLibraryItemsFunc handler (searchLibraryItemsHandler), it uses the existing search function that have been implemented. If searchLibraryItemsHandler is set, then it will be used to get search results instead.

Basically this provides a way to override the existing search function.

src/entry-point.tsx
```diff
@@ -25,6 +25,10 @@ interface RefreshLibraryViewFunc {
     (): void;
 }
 
+interface SearchLibraryItemsFunc {
+    (text: string, callback: Function): void;
```
sharadkjaiswal: Can we define an interface for the callback to be certain about the function prototype?

Randy-Ma:
@sharadkjaiswal 
As per discussion, I am merging this pull request.

PR #83, opened 2017-05-31
Fix jest testcases that fail due to api change

fanwgwg:
@monikaprabhu @Benglin


Benglin:
Thanks so much for fixing it, @fanwgwg. This is LGTM, once conflicts are resolved it'll get in.

fanwgwg:
@Benglin I've just resolved the conflicts

PR #85, opened 2017-06-01
Add lines for indentation

fanwgwg:
[*DYN-869 In tree view of Library, we should have the connecting Lines for drop down arrows.*](https://jira.autodesk.com/browse/DYN-869)

capture2

### Reviewer
@Benglin 

### FYI
@riteshchandawar


Racel:
Nice @fanwgwg!. Can you make the arrows a bit smaller?

fanwgwg:
@Racel I've updated the arrow size. Does this size look okay?

capture2

Racel:
Awesome! looks good @fanwgwg

fanwgwg:
@ellensi Can you help review this PR? Thanks!

src/resources/LibraryStyles.css
```diff
@@ -154,13 +153,18 @@ input {
...
+
+.BodyIndentationWithLine,
+.BodyIndentationWithoutLine {
+    margin-bottom: 18px;
+    padding-left: 18px;
 }
 
-.BodyIndentation {
-    padding: 8px;
+.BodyIndentationWithLine {
```
ellensi: Hi @fanwgwg,

Newly introduced elements such as `.BodyIndentationWithLine` and `.BodyIndentationWithoutLine` are actually just margins and do not need to be separate elements.

We can assign multiple class names in React:
```ts
className={LibraryItemBodyContainer}                     /* BodyIndentationWithoutLine */
className={LibraryItemBodyContainer HasExpanderLine}        /* BodyIndentationWithLine */
```

```css
.HasExpanderLine {
    margin-left: 18px;
}

.HasExpanderLine::before {
    content: "";
    border-right: 1px solid #808080;
    margin-bottom: 18px;
}
```

riteshchandawar:
@Racel In old UI we were having some gap between arrows and lines, I think that looks better than the one which we implemented above. What do you say?

image

Racel:
@riteshchandawar - I agree with you, the slight space makes it a bit clearer that the arrows are also buttons. Can we try it out and see?

src/components/LibraryItem.tsx
```diff
@@ -139,17 +139,31 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         let header: JSX.Element = null;
```
Benglin: The `render` method is going out of control, can you refactor it into multiple smaller render methods?

```js
render() {
    if (!this.props.data.visible) {
        return null;
    }

    let iconElement: JSX.Element = renderItemIcon();

    let nestedElements: JSX.Element = null;
    let clusteredElements: JSX.Element = null;
    if (this.state.expanded && this.props.data.childItems.length > 0) {
        // Same stuff goes here...
    }

    let header: JSX.Element = renderItemHeader();
    let bodyIndentation: JSX.Element = renderBodyIndentation();
    let arrow: JSX.Element = renderItemArrow();

    return (existing stuff...);
}
```

fanwgwg:
@Racel @riteshchandawar @ellensi  

I've increased the space. Does this look fine?

capture3

Racel:
@fanwgwg - space looks good, but arrow size looks maybe a pixel or 2 too small. Can you increase it a little?

fanwgwg:
@Racel  I've increased the arrow size a little bit. Is this size okay?

capture3

src/resources/LibraryStyles.css
fanwgwg: I've removed the body indentation line, but it's good to know this. Thanks!

Racel:
Looks good @fanwgwg

PR #87, opened 2017-06-05
Raise request for Library View Refresh

sharadkjaiswal:
### Purpose

This PR provides a mechanism to register request handler and raise request on library controller. When the library container doesn't have generated section, it may request for view refresh so that clients can update the `layoutSpec` as well as `loadedTypeData` and finally call `refreshLibraryView` on the library controller. This is required when ReactJS refresh the library container control [for some reason such as switching tabs in PMUI](https://github.com/DynamoDS/Dynamo/commit/eab2db8e56d8a020992df136eedb9177e358baeb), then it creates a new instance of LibraryContainer and at this moment it doesn't have the sections generated. If a client has registered a handler to serve the refresh request then the library view is refreshed properly.

### Reviewers
@Benglin

### DIFF
--- a/src/EventHandler.ts
+++ b/src/EventHandler.ts
@@ -23,7 +23,7 @@ export class Event {
     executeCallback(params?: any | any[]) {
         this.callbacks.forEach(callback => {
             try {
-                if (callback.length == 0) callback();
+                if (params.length == 0) callback();
                 else callback(params);
             }
             catch (e) {
--- a/src/components/LibraryContainer.tsx
+++ b/src/components/LibraryContainer.tsx
@@ -194,6 +194,7 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
 
     render() {
         if (!this.generatedSections) {
+            this.props.libraryController.request(this.props.libraryController.RefreshLibraryViewRequestName, function() {});
             return (<div>This is LibraryContainer</div>);
         }
 
--- a/src/components/LibraryContainer.tsx
+++ b/src/components/LibraryContainer.tsx
@@ -194,7 +194,8 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
 
     render() {
         if (!this.generatedSections) {
-            this.props.libraryController.request(this.props.libraryController.RefreshLibraryViewRequestName, function() {});
+            let eventName = this.props.libraryController.RefreshLibraryViewRequestName;
+            this.props.libraryController.request(eventName, null);
             return (<div>This is LibraryContainer</div>);
         }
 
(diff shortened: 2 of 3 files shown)


src/EventHandler.ts
```diff
@@ -23,7 +23,7 @@ export class Event {
     executeCallback(params?: any | any[]) {
         this.callbacks.forEach(callback => {
             try {
-                if (callback.length == 0) callback();
+                if (params.length == 0) callback();
```
Benglin: Wow, good catch!

src/entry-point.tsx
```diff
@@ -60,6 +64,52 @@ export class LibraryController {
         this.reactor.raiseEvent(name, params);
     }
 
+    /**
+     * This method registers the callback function when the value of a given 
+     * named variable is requested. See PackageController.request method for 
+     * more information on each available variable name.
```
Benglin: This comment isn't accurate due to the reference to `PackageController`.

src/entry-point.tsx
```diff
@@ -60,6 +64,52 @@ export class LibraryController {
...
+     * subscribed handler for this named variable.
+     */
+    request(variableName: string, callback: Function, ...argsArray: any[]) {
+
+        // Invokve handler if there's one registered.
+        let requestHandler = this.requestHandler[variableName];
+        if (requestHandler && callback) {
+            let value = requestHandler(argsArray);
+            callback(value);
+        }
```
Benglin: Ah now I see why `callback` cannot be `null`. I think it should really be just like this (allowing callers to supply `null` for `callback` if they do not care about return value):

```js
// If the a request handler was not specified, the callback 
// function will still be invoked with a 'null' as result.
let value = null; 
if (requestHandler) {
    value = requestHandler(argsArray);
}
if (callback) {
    callback(value);
}
```

src/components/LibraryContainer.tsx
```diff
@@ -194,6 +194,7 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
 
     render() {
         if (!this.generatedSections) {
+            this.props.libraryController.request(this.props.libraryController.RefreshLibraryViewRequestName, function() {});
```
Benglin: The function should be `null` if we change the implementation of `request` method as suggested towards the end. And I personally don't like lengthy lines of codes 😄 

```js
let eventName = this.props.libraryController.RefreshLibraryViewRequestName;
this.props.libraryController.request(eventName, null);
```

Benglin:
@sharadkjaiswal, would you like to update this and merge it in soon as that's done?

PR #88, opened 2017-06-09
Optimize the contruction of library

fanwgwg:
#### Previous implementation
After receiving `loadedTypes` and `layoutSpecs` and convert them to an array of `TypeListNodes` and `LayoutElement`, for each `layoutElement`, we will go to `typeListNodes` and get nodes that should go under this `layoutElement`, and build item from this. After building all the items, nodes that are not processed will go under `misc` section. For sections other than `default` and `misc`, we will remove the prefix like `pkg://` if there are any.

#### Current implementation
After receiving `loadedTypes` and `layoutSpecs` and convert them to an array of `TypeListNodes` and `LayoutElement`, we will construct items directly from `layoutElements`, just that it doesn't contain any leaf items. While constructing the items, the `parentItem` of each `includeInfo` will point to the current item. Then we will get all the `includeInfo` from the `layoutElements`, and sort them in lexicographical order by their path. Also, we will sort `typeListNodes` in lexicographical order by their `fullyQualifiedNames`. With this two ordered list, in one iteration through all the `includeInfo`, we can append all the leaf nodes to their corresponding parent items, and in this function, prefix like `pkg://` will be removed, no matter which section they should belong to. Finally, all nodes that are not processed will be grouped under `misc` section.

### Reviewer
@sharadkjaiswal


src/LibraryUtilities.ts
```diff
@@ -566,48 +281,194 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+    for (let loadedType of loadedTypes.loadedTypes) {
+        typeListNodes.push(new TypeListNode(loadedType));
+    }
+
+    for (let section of layoutSpecs.sections) {
+        sectionElements.push(new LayoutElement(section));
+    }
+
+    let sections = convertLayoutElementToItemData(sectionElements);
+    let includeInfo = getIncludeInfo(sectionElements);
```
sharadkjaiswal: includeInfo sounds like a single value, you can change the name to includeInfoList.

fanwgwg:
@sharadkjaiswal I've updated the code to allow multiple includeInfo with the same path, or one included by another, like `a.b` and `a`, `test://` and `test://a`, and validation will only return false if they have the same parent item. Also, some testcases are added for them.

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,235 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+    let previousStartingIndex = 0;
+
+    // currentStartingIndex keeps track of the index of typeListNodes when current IncludeInfo starts comparison.
+    let currentStartingIndex = 0;
+
+    for (let i = 0; i < includeItemPairs.length; i++) {
+
+        let include = includeItemPairs[i].include;
+        let includeParts = include.path.split(".");
+        let previousIncludeParts = previousIncludePath.split(".");
```
sharadkjaiswal: why not store previousIncludeParts instead of previousIncludePath?

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,235 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+
+                buildLibraryItemsFromName(node, parentItem, newName, include.iconUrl);
+            } else if (node.fullyQualifiedName.localeCompare(include.path) < 0) {
+                i--;
+                t++;
+                continue;
+            }
+        } else if (node.fullyQualifiedName.startsWith(include.path)) {
+            nodeInculde = true;
+            let prefixIndex = node.fullyQualifiedName.indexOf(prefix);
```
sharadkjaiswal: I think prefix handling can be done at the beginning and rest of the code should be same.

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,235 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+
+    // Misc section will take all unprocessed nodes
+    let miscSection = sections.find(section => section.text == miscSectionStr);
+    let unprocessedNodes = sortedTypeListNodes.filter(node => !node.processed);
+    unprocessedNodes.forEach(node => buildLibraryItemsFromName(node, miscSection));
+
+    for (let section of sections) {
+        // Change the itemType of the outermost nodes to category
+        section.childItems.forEach(item => {
+            if (item.itemType == "group") {
```
sharadkjaiswal: Why do we bother testing the itemType, we can directly assign if the second level items are always of the category type?

src/LibraryUtilities.ts
```diff
@@ -666,20 +571,44 @@ export function buildLibraryItemsFromName(typeListNode: TypeListNode, parentNode
     // Otherwise, create the new parent node 'A' (using the previous example).
     let newParentNode = new ItemData(fullyQualifiedNameParts[0]);
     newParentNode.constructFromTypeListNode(typeListNode);
-    newParentNode.itemType = "group";
+
+    if (iconUrl) {
+        newParentNode.itemType = "none";
```
sharadkjaiswal: Hmm, so can we not skip setting itemType on spec elements?

fanwgwg: Nope we are setting it to `none` or `group`

fanwgwg:
@sharadkjaiswal I've updated the implementation and testcases.

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+        if (parentItem) {
+            parentItem.appendChild(layoutData);
+        }
+
+        if (layoutElement.include.length > 0) {
+            for (let include of layoutElement.include) {
+                let pair = new IncludeItemPair();
+                pair.include = include;
+                pair.parentItem = layoutData;
+                pairs.push(pair);
```
Benglin: Let's update constructor of `IncludeItemPair` so that the following is possible:

```js
pairs.push(new IncludeItemPair(include, layoutData));
```

Benglin: Then you can use `forEach` or something like that.

src/LibraryUtilities.ts
```diff
@@ -38,6 +38,11 @@ export interface IncludeInfo {
     inclusive?: boolean;
 }
 
+export class IncludeItemPair {
+    include: IncludeInfo;
+    parentItem: ItemData;
+}
```
Benglin: The term *parent* gives a confusing meaning in hierarchical relationship, I didn't understand this until I read the code bellow, and now I think a better name would be `owningItem`. Does that make sense?

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
 }
 
 /**
- * Convert an array of typeListNodes to ItemData which are not processed based on their fullyQualifiedNames, 
- * by splitting the name with '.', and put them under a specific section.
+ * This function will convert layoutElements to ItemData, and construct an array of IncludeItemPair,
+ * which stores an includeInfo and its parent item.
```
Benglin:
```js
/**
 * This function converts LayoutElement objects to corresponding ItemData objects 
 * in hierarchical structure. As part of this process, the function generates a 
 * list of IncludeItemPair objects for each of the IncludeInfo object under a given 
 * LayoutElement object. 
 */
```

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
  * 
- * As an example, for two typeListNodes 'A.B.C' and 'A.B.D', the resulting nodes are:
- * - A
- *   |- B
+ * @param {LayoutElement[]} layoutElements 
+ * The layoutElements to be converted to ItemData
```
Benglin: Since we are referring to the objects by its type, then a capital `LayoutElement` should be used:

`The list of LayoutElement objects to be converted to ItemData`

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+ * 
+ * @param {IncludeItemPair[]} pairs
+ * The IncludeItemPairs to be contructed from
```
Benglin:
```js
// A list of IncludeItemPair objects that are generated, each for an 
// IncludeInfo under a given LayoutElement object.
```

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+ * 
+ * @param {ItemData} parentItem 
+ * The parentItem of layoutElements (if exist)
```
Benglin:
```js
// The parent LayoutElement object for the current list of LayoutElement objects
// (layoutElements) that are being processed, if one exists.
```

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
+ * - parentItem
+ *  |- B
  *      |- C
- *      |- D
+ *          |- D
  * 
- * @param {TypeListNode[]} allNodes 
- * All the typeListNodes
+ * If there is not a match or include, skip to next includeInfo/typeListNode, depending on their 
+ * lexicographical order.
```
Benglin: This is really a good example of great documentation, keep up the good work, @fanwgwg 👍 I would as usual suggest some tweaks to the documentation as follow:

/**
 * This function merges all TypeListNode objects under the corresponding ItemData 
 * (i.e. a library item) based on information provided in IncludeItemPair. Each 
 * IncludeItemPair stores an IncludeInfo and the owning ItemData for that IncludeInfo.
 * 
 * When the function iterates through each IncludeItemPair object, comparison happens 
 * between fullyQualifiedName of a given TypeListNode, and the IncludeInfo.path found 
 * in each IncludeItemPair.
 * 
 * If fullyQualifiedName matches IncludeInfo.path, it means they are exactly the same 
 * and therefore the TypeListNode will be merged to the owning ItemData specified in 
 * the IncludeItemPair.
 * 
 * If IncludeInfo.path does not completely match fullyQualifiedName but represents a 
 * substring of fullyQualifiedName, then the following takes place:
 * 
 * Hierarchical structure of ItemData objects will be generated by looking at the 
 * fullyQualifiedName of TypeListNode. These ItemData objects will then be merged 
 * under the owning ItemData specified in the IncludeItemPair. For example: if the 
 * fullyQualifiedName is "A.B.C.D" and IncludeInfo.path is "A.B", then the generated 
 * ItemData take the following form:
 * 
 * - owningItem
 *    |- B
 *       |- C
 *          |- D
 * 
 * If there is no match between fullyQualifiedName and IncludeInfo.path, then skip to 
 * the next IncludeInfo/TypeListNode depending on their lexicographical order.
*/

Benglin: Please don't forget to update this comment after you've renamed `IncludeItemPair` later.

src/LibraryUtilities.ts
```diff
@@ -566,48 +285,244 @@ export function updateSections(oldLayoutSpecs: any, newLayoutSpecs: any): void {
...
  */
-export function convertToMiscSection(allNodes: TypeListNode[], section: LayoutElement): ItemData {
+export function constructFromIncludeInfo(typeListNodes: TypeListNode[], includeItemPairs: IncludeItemPair[]) {
+    let t = 0;
+    let prefix = "://";
+    let nodeMatch = false;
+    let nodeInculde = false;
+
+    // previousIncludeParts is the includeInfo path parts of the pair before current pair.
+    let previousIncludeParts: string[] = [];
```
Benglin: I could not follow this function, I will let this in, and debug it to see what it does, and then add necessary comments.

Benglin:
Hi @fanwgwg, again thanks for the great work and @sharadkjaiswal for the guidance. I understand the intents of this PR and have spent most of my reviewing time understanding the specifics. The test cases are great and I'm confident that they will go a long way in validating future code changes. I think if Sharad does not have further comments I would like to get this in, and then debug/update comments as I go. @fanwgwg, please drop by and get me to merge this for you later. Thanks

PR #92, opened 2017-06-19
Remove unused field that has been exposed from SearchBar

Randy-Ma:
### Purpose

The function interface SetSearchInputFieldFunc served to expose an internal field searchInputField and the member searchInputField in the class Searcher are not used anywhere any more. 

This is to remove them.

### Reviewers
@Benglin


Benglin:
Thanks @Randy-Ma! LGTM

PR #93, opened 2017-06-19
Restore command for testing utility functions

fanwgwg:
### Reviewer
@Benglin

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,12 @@ Installing all dependencies
 
 ## Run Tests
 
-- Run all tests
+####  Run all tests
 
     `$ npm run test`
 
     Test file is at folder `./__tests__/`
-
-- Run test for utility functions
+####  Run test for utility functions
 
     `$ npm run utiltest`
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,9 +22,17 @@ Installing all dependencies
...
 ## Run Tests
 
+- Run all tests
+
     `$ npm run test`
 
-    Test file is at folder `./__test__/`
+    Test file is at folder `./__tests__/`
+
+- Run test for utility functions
```
Benglin: I find it odd that my previous comments four days ago didn't show up here anymore. But IIRC, the comments went something like this:

Do make the following changes:

- `Run all tests` and `Run test for utility functions` should not be bullet points, they can just be a header `#### Run all test` for example
- Under `Run test for utility functions` section, add the following comments (to explain why is this even required):

    > `Run all tests` command does run all the tests, but it does not output `libraryUtilitiesTest.ts` console messages. When any test fail, it is unclear which one does and why. When that happens, use the command below to run just the utility tests.

PR #94, opened 2017-06-19
Fix a bug in function removeEmptyNodes and update testcases

fanwgwg:
Function `removeEmptyNodes` will take in an array of `ItemData`, and remove all nodes that have no child items (including grandchild items).

### Reviewer
@Randy-Ma

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -515,7 +515,7 @@ export function removeEmptyNodes(items: ItemData[]) {
     items.forEach((item, index, items) => {
         if (item.childItems.length > 0) {
             if (removeEmptyNodes(item.childItems)) {
-                removeEmptyNodes(items);
+                removeEmptyNodes([item]);
             }
         } else if (item.itemType === "section" || item.itemType === "category" || item.itemType === "group") {
             items.splice(index, 1);
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -512,17 +512,18 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
 export function removeEmptyNodes(items: ItemData[]) {
     let itemRemoved = false;
 
-    items.forEach((item, index, items) => {
+    for (let i = 0; i < items.length; i++) {
+        let item = items[i];
         if (item.childItems.length > 0) {
             if (removeEmptyNodes(item.childItems)) {
-                removeEmptyNodes([item]);
+                i--;
             }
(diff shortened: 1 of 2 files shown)


src/LibraryUtilities.ts
```diff
@@ -510,13 +510,21 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
 
 // Remove empty non-leaf nodes from items
 export function removeEmptyNodes(items: ItemData[]) {
+    let itemRemoved = false;
+
     items.forEach((item, index, items) => {
         if (item.childItems.length > 0) {
-            removeEmptyNodes(item.childItems);
+            if (removeEmptyNodes(item.childItems)) {
+                removeEmptyNodes(items);
```
Randy-Ma: Unnecessary computation is happening here. Assuming there are 1000 items, the last one has some child items removed, then all the first 999 items will be tried again.

src/LibraryUtilities.ts
```diff
@@ -510,13 +510,21 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
 
 // Remove empty non-leaf nodes from items
 export function removeEmptyNodes(items: ItemData[]) {
+    let itemRemoved = false;
+
     items.forEach((item, index, items) => {
         if (item.childItems.length > 0) {
-            removeEmptyNodes(item.childItems);
+            if (removeEmptyNodes(item.childItems)) {
+                removeEmptyNodes([item]);
```
Randy-Ma: I am pretty sure that this won't actually remove item from the original items array.

fanwgwg: This is actually still passing by reference, and I've tried and tested that it works.

src/LibraryUtilities.ts
```diff
@@ -510,13 +510,21 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
...
+            }
         } else if (item.itemType === "section" || item.itemType === "category" || item.itemType === "group") {
             items.splice(index, 1);
+            itemRemoved = true;
+            removeEmptyNodes(items);
```
Randy-Ma: If we remove this line, it does not work?

src/LibraryUtilities.ts
Randy-Ma: Hmmmm

Randy-Ma: I mean item will be removed from the new array - [item], but not from the original array.

PR #97, opened 2017-06-20
Small tweaks to improve developer experience when importing librarie.js

pboyer:


src/components/LibraryContainer.tsx
```diff
@@ -20,12 +20,12 @@ export interface LibraryContainerProps {
 }
 
 export interface LibraryContainerStates {
-    inSearchMode: boolean,
-    searchText: string,
-    selectedCategories: string[],
-    structured: boolean,
-    detailed: boolean
-    showItemSummary: boolean
+    inSearchMode?: boolean,
```
mjkkirschner: @pboyer can you splain this pattern to me? 
If all the properties of these interfaces are optional, then shouldn't the code also reflect that and be fault tolerant to  missing properties? Is it?

package.json
```diff
@@ -1,6 +1,7 @@
 {
-  "name": "dynamo-hosted-contents",
-  "description": "Project that contains all hosted contents of Dynamo Windows client",
+  "name": "librarie.js",
+  "version": "0.0.1",
+  "description": "Reusable lightweight library component based on React.js",
```
Benglin: Oh wow thanks @pboyer !

src/components/LibraryContainer.tsx
Benglin: Yah, not entirely sure why are these changes made, @pboyer, there's no [documentation](https://github.com/DynamoDS/Dynamo/wiki/Incremental-Pull-Request-Template) on the PR so you need to buy us donuts!

PR #98, opened 2017-06-22
Add layoutSpecs for Revit nodes

fanwgwg:
capture

### Reviewer
@sharadkjaiswal


Benglin:
Thanks @fanwgwg I know this involves a lot of manual work, but you got it done. Let me just merge it in so we can try things out on Revit.

PR #101, opened 2017-09-18
Update LayoutSpec

QilongTang:
Include `RevitNodes.Revit.Elements.Views.AreaPlanView` in Views category

FYI: @mjkkirschner @ramramps @alfarok @ColinDayOrg @Racel @smangarole @jnealb @sm6srw @saintentropy 

From current set-up, and node added by a third-party PR, we will need to touch this file manually. Otherwise, node will fall into wrong category:
image

After this change is merged, one need to build Librarie.js locally and update the min.js in DynamoProject so that this change can be consumed.
image

### DIFF
--- a/docs/LayoutSpecs.json
+++ b/docs/LayoutSpecs.json
@@ -1838,6 +1838,9 @@
                 "elementType": "group",
                 "include": [
                   {
+                    "path": "RevitNodes.Revit.Elements.Views.AreaPlanView"
+                  },
+                  {
                     "path": "RevitNodes.Revit.Elements.Views.AxonometricView"
                   },
                   {


mjkkirschner:
@QilongTang are you sure about this? The DynamoRevit repo has its own layout spec doesn't it?

ramramps:
@QilongTang Can you update the Layoutspec in Dynamo and see whether librarie.min.js takes the latest changes?  my guess is that layoutspec is not integrated in librarie.min.js. So, if you update layoutspec in Dynamo, it should display the correct values.

mjkkirschner:
why should the Librarie.js repo know anything about dynamo or dynamo revit?

ramramps:
yes @mjkkirschner. layoutspec's are different. Librarie.min.js should read the layoutspec in the current directory.  So, every source can have its own layoutspec.

QilongTang:
@mjkkirschner @ramramps I'n now sure about this :) The layoutspecs.json here in this repo is split into Dynamo Part and none Dynamo Part (which is the Revit Cate alone), the json here is source of truth. Any update here, there should be an update reflecting the change in Dynamo or DynamoRevit Repo.

ramramps:
@QilongTang Yes. We need to change that. LayoutSpec should not live inside here. It should be in S3. Build should pull from S3, when creating the package. If no LayoutSpec is specified, it should use the one that comes in the package. I will document this one of the librarie.js task.

mjkkirschner:
huh, why is s3 involved?

QilongTang:
@ramramps @mjkkirschner  Let's not rush into any decision without thinking for the 3rd party dev aspect. PR for DynamoRevit repo: https://github.com/DynamoDS/DynamoRevit/pull/1796. Will cherry-pick to other branches.

mjkkirschner:
I agree, not only 3rd party devs, but other consumers of Dynamo as a core component, they need to be able to layout their libraries without support from us.

ramramps:
S3 is involved to store the layoutspec.json files for the suite we support.

PR #102, opened 2017-11-07
Scroll expanded items into view

mjkkirschner:
nov-07-2017 14-51-47

#### Purpose:
This PR adds a new method to the `libraryContainer` which attempts to scroll to a certain `LibraryItem` while keeping its position on the screen the same before and after an expansion of that node. This is an issue because when we open a new subset of the library we close all others.

I have added some tests that assert the methods are raised when clicks occur, but I was not able to actually test the scrolling as the elements bounding rectangles show them both overlapping even when using mount with enzyme - unless I am missing something we might be able to test this with selenium in the future when integrated into AVP.

The scroll method is passed down to nested libraryItems at construction time within their `willExpand` callback

I've also updated the snapshot tests to match the current behavior (these were out of date before I started this PR)


mjkkirschner:
@ramramps @QilongTang @alfarok - please review

src/components/LibraryContainer.tsx
```diff
@@ -98,6 +101,47 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
+     * This method attempts to scroll the outer libraryItemContainer so that the position
+     * on screen of the passed element does not change, even though the scroll position does.
+     * @param element 
+     */
+    scrollToExpandedItem(element: HTMLElement) {
+        if (element) {
+            var currentElement = ReactDOM.findDOMNode(this).querySelector(".LibraryItemContainer");
+            var currentDocScrollTop = currentElement.scrollTop;
+
+            console.log(currentDocScrollTop);
```
ramramps: remove this please

src/components/LibraryContainer.tsx
```diff
@@ -98,6 +101,47 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
+                // this should be the current top position of the element after it's been expanded
+                var currentTopOfElementAfterExpand = this.offset(element).top;
+                //this is the new scroll position which will make it appear as if
+                //the element is in the same location on screen, even though the scroll
+                //has changed.
+                var newOffset = currentTopOfElementAfterExpand - oldOffset;
+
+                currentElement.scrollTop = newOffset;
+
+            }, 1);
```
ramramps: did you try with 0 instead of 1?

mjkkirschner: not yet, will give it a shot.

alfarok:
@mjkkirschner also seeing the jumping in Import/Export -> FileSystem

library_scroll

mjkkirschner:
@alfarok the gif you posted above does not replicate the issue since when you open image - the window is not small enough or contain enough content any longer to put image in the same position.

mjkkirschner:
@alfarok but I have verified it does occur for units and operators in math, so will look into that. I think it has to do with different nesting level where I need to apply the scroll behavior in the other constructors for libraryItem at all levels.

alfarok:
Ahh I see what you mean, let me know if you need me to test it again with any changes

mjkkirschner:
@QilongTang I've given up on figuring out the right calculations for each of the different edge cases, and instead just do a simple march by 1 pixel to attempt to scroll to the correct position based on the previous position of the element. This works well and is fast enough - seems just as fast as perform in my tests - (though obviously orders of magnitude slower than just doing a few subtractions... We could make this more performant in the future for very large categories by doing an adaptive march or just using binary search.

It does bail out if it must traverse more than 10,000 pixels or if theres no possible way to scroll to the correct position (not enough scrolling space).

If anyone has other ideas would be glad to hear them, but this seems to reduce code, I could not identify the correct calculations for all cases of growing and shrinking the library - though it must be possible.

mjkkirschner:
@QilongTang @alfarok 
after staring at that loop for a while 🌽  - pun intended!

I noticed that it we should be able to just move the scroll by the distance between the old position and the new position - This works atleast for the initial case and the case that @alfarok found, no scary while loop, and only one case.

QilongTang:
@Racel As you requested, here is the snapshot of multiple categories expanding at same time. If you feel like  node items should be smaller, we can go ahead with @mjkkirschner 's approach and deal with this later not as part of Dynamo 2.0. Othewise, we might want to change where we are heading
image

__tests__/UITests.tsx
```diff
@@ -104,15 +104,64 @@ describe("LibraryContainer UI", function () {
...
+    // Load the data to libController
+    libController.setLoadedTypesJson(loadedTypesJson, false);
+    libController.setLayoutSpecsJson(layoutSpecsJson, false);
+    libController.refreshLibraryView();
+
+    let header = libContainer.find('div.LibraryItemHeader').at(0);
+    chai.expect(header).to.have.lengthOf(1);
+    var count = 0;
+    //replace the scroll method with a method which ends the test.
+    libContainer.getNode().scrollToExpandedItem = () => { count = count + 1; }
```
QilongTang: Is it because we do not have sinon or similar dependency which can mock this for us?

mjkkirschner: I could attempt it that way, I think sinon or something with a spy API is loaded, I just prefer this for simplicity and a new entity is built for each test so it should not effect other tests. Can try to use a spy if you prefer.

mjkkirschner: @QilongTang I lied, I don't see any use of spies in the repo with a quick search, must have been somewhere else.

QilongTang: OK. I don't mean that we need to introduce that as part of this PR, was just curious

QilongTang:
LGTM

PR #104, opened 2017-11-14
Remove structured view

mjkkirschner:
This PR removes the structured search view button - it leaves all code in place and another task will be created for removing this dead code.

A UI test is also added that renders the dom and asserts the correct number of buttons exist before and after searching.

### DIFF
--- a/__tests__/UITests.tsx
+++ b/__tests__/UITests.tsx
@@ -261,4 +261,31 @@ describe("LibraryContainer UI", function () {
       done(); // For testframework to know when to terminate execution
     }, 500);
   });
+
+  it("search bar should not contain structured view button", function () {
+    
+        
+        let libContainer = mount(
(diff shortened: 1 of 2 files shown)


mjkkirschner:
please review @ramramps @QilongTang

ramramps:
what it takes to remove the dead code?

mjkkirschner:
@ramramps it flows down to most parts of the codebase in library item, search etc and is used to make decisions based on structure state true or false. It's not too bad to remove but will touch a bunch of files.

ramramps:
Keeping that dead code is fine. Might be useful if we want to get this back

mjkkirschner:
@QilongTang @ramramps as soon as this is good to merge I will update library min in dynamo.

QilongTang:
LGTM

PR #105, opened 2017-11-22
Fix for item expand

ramramps:
This PR fixes the issue with `ItemToExpand`.

As per Librarie.js code :
 If the item is defined in layoutspec and not in Rawdatatype, it will not be there in the search result.
 If the item is defined in RawDataType and not in layoutspec, it will be in miscallenous section.

So, searching for an item which is not there in RawDataType.json, will not fetch any thing in search view. `findAndExpandItemByPath` is called only when the search returns the items.

So, search is returning the item, and doing item.expand == true will make the nested levels to expand correctly.  So, checking for item to be there in allItems is irrelevant here . 

@QilongTang @ColinDayOrg @mjkkirschner @smangarole

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -747,7 +747,19 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
  * @return {boolean} true if an item is found, false otherwise
  */
 export function findAndExpandItemByPath(pathToItem: ItemData[], allItems: ItemData[]): boolean {
-    let item: ItemData;
+
+    //If RawDataType is handcrafted, then there is a chance for the item
+    //to get incorrect information. Otherwise, pathToItem has the right members
+    //just expanding the item to true will work. 
+    pathToItem.forEach(element => {
+        element.expanded = true;
+    });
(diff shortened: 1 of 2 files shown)


ColinDayOrg:
LGTM

src/LibraryUtilities.ts
```diff
@@ -747,7 +747,19 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
...
+    //just expanding the item to true will work. 
+    pathToItem.forEach(element => {
+        element.expanded = true;
+    });
+
+    return true;
+   
+    //if the above code did not work, then uncomment the below
+    //code to find the actual issue.
+   /* let item: ItemData;
```
QilongTang: Did you find out any case you will need to uncomment this below yet?

ramramps: not yet. still testing

ramramps:
Is it ok to merge this and check D4R? I believe this doesn't hurt other behavior in librarie.js

ramramps:
Merging this.

PR #106, opened 2017-11-25
LibrarieJS fix Item click code update

ramramps:
The previous PR had the fix which is very slow in D4R and Dynamo Sandbox. So, reverting the code with removed check for icon url. This is much faster than the previous approach

@QilongTang

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -748,21 +748,10 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
  */
 export function findAndExpandItemByPath(pathToItem: ItemData[], allItems: ItemData[]): boolean {
 
-    //If RawDataType is handcrafted, then there is a chance for the item
-    //to get incorrect information. Otherwise, pathToItem has the right members
-    //just expanding the item to true will work. 
-    pathToItem.forEach(element => {
-        element.expanded = true;
-    });
-
-    return true;
-   
-    //if the above code did not work, then uncomment the below
-    //code to find the actual issue.
-   /* let item: ItemData;
-
+   let item: ItemData;
+    //commented the check for iconUrl. This is false only for static RawDataType files.
     item = allItems.find(item =>
-        item.text == pathToItem[0].text && item.iconUrl == pathToItem[0].iconUrl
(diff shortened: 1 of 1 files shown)


src/LibraryUtilities.ts
```diff
@@ -748,21 +748,10 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
  */
 export function findAndExpandItemByPath(pathToItem: ItemData[], allItems: ItemData[]): boolean {
 
-    //If RawDataType is handcrafted, then there is a chance for the item
-    //to get incorrect information. Otherwise, pathToItem has the right members
-    //just expanding the item to true will work. 
-    pathToItem.forEach(element => {
```
ramramps: this is very slow for a huge number of items :(

src/LibraryUtilities.ts
```diff
@@ -748,21 +748,10 @@ export function getHighlightedText(text: string, highlightedText: string, matchD
...
-   
-    //if the above code did not work, then uncomment the below
-    //code to find the actual issue.
-   /* let item: ItemData;
-
+   let item: ItemData;
+    //commented the check for iconUrl. This is false only for static RawDataType files.
     item = allItems.find(item =>
-        item.text == pathToItem[0].text && item.iconUrl == pathToItem[0].iconUrl
+        item.text == pathToItem[0].text //&& item.iconUrl == pathToItem[0].iconUrl
```
ramramps: commented this. this check is not required for dynamic data files.

QilongTang: can we remove this instead?

ramramps: we can keep it for sometime (until the dynamo, d4r testing is done).

QilongTang:
library

From the testing, I found that clicking on the category only collapse other cates and expand that corresponding one but not necessary scroll to that one which is not the same with the built in categories. Can we align these two behavior?

ramramps:
@tanga I did not fix this issue. The task says it should be a separate task.

QilongTang:
@ramramps The first AC says -> Clicking on the 'Element' category link under Element.Geometry should bring the view back to the main view of the library, expand the 'Element' subcategory, and scroll to it. 
@Racel Is it fine for not including the scrolling behavior as part of this task?

PR #108, opened 2017-12-04
Remove icons from Geometry class sub-categories

alfarok:
### Purpose

[QNTM-2675](https://jira.autodesk.com/browse/QNTM-2675)

[Supplemental Dynamo PR](https://github.com/DynamoDS/Dynamo/pull/8369)

The purpose of this PR is to remove the icons from the Geometry class sub-categories.  The icons beside the Geometry categories is confusing as they look like nodes and not sub-categories.

There are several element types in library which determine the properties each layer has.  The`none` element type contains all expandable library items that are not `categories` or `groups` (essentially all the interim dropdowns).  With this PR even if an icon url is specified it will be ignored so icons can now only be added at the top and bottom levels of the library hierarchy (`categories` and nodes `create` `action` `query`)

For more info on Element Types see the [Librarie Docs
](https://github.com/DynamoDS/librarie.js/blob/master/docs/layout-specs.md)
image

### Declarations

- [X] The code base is in a better state after this PR
- [X] Is documented according to the [standards](https://github.com/DynamoDS/Dynamo/wiki/Coding-Standards)
- [X] The level of testing this PR includes is appropriate
- [X] User facing strings, if any, are extracted into `*.resx` files
- [X] All tests pass using the self-service CI.
- [X] Snapshot of UI changes, if any.
- [X] Changes to the API follow [Semantic Versioning](https://github.com/DynamoDS/Dynamo/wiki/Dynamo-Versions), and are documented in the [API Changes](https://github.com/DynamoDS/Dynamo/wiki/API-Changes) document.

### Reviewers

@mjkkirschner 

### FYIs

@Racel

### DIFF
--- a/__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
+++ b/__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
@@ -146,11 +146,6 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
         onError={[Function]}
         src="test-file-stub"
       />
-      <img
-        className="LibraryItemIcon"
-        onError={[Function]}
-        src=""
-      />
       <div
         className="LibraryItemText"
       >
@@ -182,11 +177,6 @@ exports[`LibraryContainer Test UI rendering of single component of Library Item 
       className="Arrow"
       onError={[Function]}
       src="test-file-stub"
-    />
-    <img
-      className="LibraryItemIcon"
-      onError={[Function]}
-      src=""
     />
     <div
       className="LibraryItemText"
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -158,8 +158,9 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     }
 
     getIconElement(): JSX.Element {
-        // Group displays only text without icon.
-        if (this.props.data.itemType !== "group" && this.props.data.itemType !== "section") {
+        let itemType = this.props.data.itemType;
+        // If the element type is a section, group, or none an icon shouldn't be displayed
+        if (itemType !== "group" && itemType !== "section" && itemType !== "none") {
             return <img
                 className={"LibraryItemIcon"}
                 src={this.props.data.iconUrl}


src/components/LibraryItem.tsx
```diff
@@ -158,8 +158,9 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     }
 
     getIconElement(): JSX.Element {
-        // Group displays only text without icon.
-        if (this.props.data.itemType !== "group" && this.props.data.itemType !== "section") {
+        let itemType = this.props.data.itemType;
+        // If the element type is a section, group, or none an icon shouldn't be displayed
+        if (itemType !== "group" && itemType !== "section" && itemType !== "none") {
```
ramramps: may be you should check the tests and see this is covered.

alfarok: @ramramps yes will do, the AC calls for a unit test

Racel:
from ux perspective, looks good to me! thanks @alfarok

mjkkirschner:
@alfarok also please check the docs to see if the iconUrl is in the layout spec somewhere - if it is, should it now be removed?

alfarok:
@mjkkirschner correct, they can be removed from the layout spec.  If any are added or left in other places they should be simply ignored.

__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
```diff
@@ -146,11 +146,6 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
         onError={[Function]}
         src="test-file-stub"
       />
-      <img
-        className="LibraryItemIcon"
-        onError={[Function]}
-        src=""
-      />
       <div
```
alfarok: @mjkkirschner @ramramps these tags are no longer generated since the `none` element type is no longer adding icons so I have to update the snapshot to exclude them.  If they were to get generated the test will now fail.  You can see the tests that generate these 2 snaps [here](https://github.com/DynamoDS/librarie.js/blob/master/__tests__/Snapshottest/UIOutputComparisonTests.tsx#L41).

alfarok: Dynamo dynamically generates the data in the libraryUI so it is hard ensure all the required items are loaded in Librarie, for example if I wanted to verify icons are still generated for the nodes.  I am not sure if this is adequate or if there is a better approach I could be taking here.

PR #109, opened 2017-12-06
LibrarieJS UI changes

ramramps:
This  PR is for LibrarieJS UI Changes.

1. Introduced "coregroup" for elements defined in Layoutspec.
2. Introduced "class" for elements defined in Layoutspec
3. Removed arrows for "coregroup"
4. Auto expand "coregroup" elements.
5. New icons

giphy 2

@mjkkirschner @alfarok @ColinDayOrg 

FYI Changes are made after review from @Racel


src/components/LibraryItem.tsx
```diff
@@ -222,6 +230,18 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
 
+        if (this.props.data.itemType == "category") {
+            if (!this.state.expanded) {
+                arrowIcon = require("../resources/ui/indent-arrow-right-wo-lines.svg");
+            }
+            else {
+                arrowIcon = require("../resources/ui/indent-arrow-down-wo-lines.svg");
+            }
+
+            return <img className={"CategoryArrow"} src={arrowIcon} onError={this.onImageLoadFail} />;
```
ramramps: for arrow alignment on category

src/components/LibraryItem.tsx
```diff
@@ -370,12 +392,22 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
         let currentlyExpanded = this.state.expanded;
+
         if (this.props.data.childItems.length > 0 && !currentlyExpanded && this.props.onItemWillExpand) {
             this.props.onItemWillExpand(ReactDOM.findDOMNode(this));
         }
 
         this.setState({ expanded: !currentlyExpanded });
 
+        //auto expand the coregroup elements
+        if(this.props.data.itemType === "category" ) {
```
mjkkirschner: does this work for nested categories?

docs/LayoutSpecs.json
```diff
@@ -808,7 +808,7 @@
             {
               "text": "Color",
               "iconUrl": "",
-              "elementType": "group",
+              "elementType": "class",
```
mjkkirschner: so when does class get used? When should it be used opposed to group? Can you add these types to the documentation files? / wiki?

Do we ever generate `classes` or `coreGroups` when importing some other library which is not defined in the layout spec?

src/components/LibraryItem.tsx
mjkkirschner: @Racel curious if this case was considered?

docs/LayoutSpecs.json
mjkkirschner: @ramramps also, so you will need to update the layoutspec in revit and dynamo as a consequence of this PR as well to get the same behavior right?

mjkkirschner: First, I understand this work is being done to accomplish the particular design requirements.
I think it would be useful for someone to enumerate the rules that are now encoded in this repo.
I could be wrong, but they seem pretty complicated to me.

I will take a shot:
* Top Level categories display Icons
* groups or classes do not display icons
* NonNested categories do not expand
* Top Level categories expand
* other stuff?

I do not think we will be able to maintain this matrix of interactions without properly documenting it and the new classes/ groups, etc...

This is not a criticism of the implementation here. I am fine with this approach, but would be happier if we did not need to introduce new types to accomplish this, just apply the minimal logic required to existing types throughout the library with no special logic for items in the layout spec - remember we'll need another giant layout spec for TS.

at a minimum let us please document these rules throughly.

mjkkirschner:
@ramramps looks good implementation wise - some questions about other approaches not requiring new types, but LGTM if you have thought about that already.

src/components/LibraryItem.tsx
Racel: @mjkkirschner Only needed for top level categories.

docs/LayoutSpecs.json
ramramps: @mjkkirschner Thanks for the comments. 
1. Dynamo and Revit layoutspec needs a change. 
2. This is definitely to get that particular design. This was easy in WPF. But in librarie.js, everything is dependent on `elementType`.  I don't have a choice other than adding a new one. 
3. Auto expand - as per the task, it is only for top level. Per my understanding from layoutspec, this is the order after the change`section-category-coregroup-group-none`. All the nested category gets `group` by default. 
4.Before the change, the order is `section-category-group-none`. So, there is no clear distinction between the top level and nested categories. Well, we can leave it this way too.
5. I am afraid that if layoutspec is auto generated, then achieving this design will get complicated and requires careful look of how it is made today.
6.I will add a wiki page with rules per my understanding. 

Also, I will rename `coregroup to groups`.  So, the top level has groups of group.

ramramps:
@mjkkirschner @Racel After giving it another shot, there is another way to achieve it. 
No updates to layoutspec is required. `group` to `coregroup` is changed on the fly now. 
So, no major updates required for layoutspec in revit or dynamo.
There is no other way for `class` update. that has to be made in layoutspec.

src/LibraryUtilities.ts
```diff
@@ -508,6 +511,37 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
...
+            return child.elementType == "category";
+        }); 
+
+        //get the first level group elements
+        let groupElements = categoryElements.filter((child: LayoutElement) => {
+            //change the element type of group elements
+            child.childElements.forEach((grp: LayoutElement) => {
+                if(grp.elementType != "class") {
+                    grp.elementType = "coregroup";
+                }
```
ramramps: I could have used underscore, but I find this way more explicit for someone to understand what we are doing (underscore does this implicit).

__tests__/UITests.tsx
```diff
@@ -330,4 +334,24 @@ describe("LibraryContainer UI", function () {
       done();
     }, 500);
   });
+
+  it("coregroup items should auto expand", function () {
```
ramramps: this test covers the major part.

__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
```diff
@@ -146,6 +146,11 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
         onError={[Function]}
         src="test-file-stub"
       />
+      <img
```
mjkkirschner: so I think you might need to update the tests file after you merge with master as I think this will undo what Keith added.

__tests__/UITests.tsx
mjkkirschner: 👍

docs/LayoutSpecs.json
mjkkirschner: @ramramps thanks for the detailed response - have you added the wiki page yet (6) on your list?

__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
ramramps: ah..I think I discarded that change. Will check the master file.

docs/LayoutSpecs.json
ramramps: not yet. the rules are simplified now. I will start wiki , and let @Racel add the rules. She knows better than me.

src/LibraryUtilities.ts
mjkkirschner: cool - looks good, can you add just a bit more description to the summary of the behavior - something like: changes categories topLevel children `itemType` from groups to coreGroups.... or something like that.

src/LibraryUtilities.ts
```diff
@@ -667,12 +701,17 @@ export function setItemStateRecursive(items: ItemData | ItemData[], visible: boo
...
+            item.expanded = true;
+        }
+
         setItemStateRecursive(item.childItems, visible, expanded);
     }
 }
 
 export function search(text: string, item: ItemData) {
-    if (item.itemType !== "group") {
+    if (item.itemType !== "group" && item.itemType !== "coregroup") {
```
mjkkirschner: should we also ignore class during search?

mjkkirschner:
@ramramps thanks for taking another look at this - just a few more comments.

src/LibraryUtilities.ts
ramramps: I don't think so. the idea is to change to class and use the search

ramramps:
@Racel https://wiki.autodesk.com/display/GEN/Rule+book
Here is a very basic one. We can update / remove contents from this.

ramramps:
@Racel @mjkkirschner Let me know whether I can merge this code. 
I will send PR for Dynamo after that.

Racel:
LGTM. Lets merge and see how it goes...

Racel:
@ramramps - Before you merge, can you remove the arrow from the very top level category? I think it is too much...

QilongTang:
@ramramps We will just merge this one and create different defect for UI improment. Thanks

src/components/LibraryItem.tsx
```diff
@@ -178,9 +182,12 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
     getIconElement(): JSX.Element {
-        let itemType = this.props.data.itemType;
-        // If the element type is a section, group, or none an icon shouldn't be displayed
-        if (itemType !== "group" && itemType !== "section" && itemType !== "none") {
+        // If the element type is a section, group, coregroup, class or none an icon shouldn't be displayed
+        if (this.props.data.itemType !== "group" && 
+        this.props.data.itemType !== "section" && 
+        this.props.data.itemType !== "coregroup" &&
+        this.props.data.itemType !== "class" &&
+        this.props.data.itemType == "none") {
```
alfarok: should be `this.props.data.itemType !== "none"`

src/resources/icons/arrow-pointing-to-right.svg
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
QilongTang: This icon does not seem referenced anywhere!

src/resources/icons/library-action-old.svg
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
QilongTang: This icon does not seem referenced anywhere!

src/components/LibraryItem.tsx
QilongTang: This should be `!==` will address in a different PR

PR #111, opened 2017-12-13
Update LibrarieJS

ramramps:
This PR is to update  `class` to  `classType`. This is because `class` is a keyword in c#.

Related DynamoPR: https://github.com/DynamoDS/Dynamo/pull/8397

@QilongTang


QilongTang:
LGTM

PR #115, opened 2018-02-20
Remove the miscellaneous section

ramramps:
This PR is to Remove the Miscellaneous section from Librarie for Dynamo.

Task : https://jira.autodesk.com/browse/QNTM-2976

1. commented the code that adds nodes to Miscellaneous section. May be we might need it in future.
2. Layoutspec in LibrarieJS has Miscellaneous section. I did not remove that part (as this is more general and not related to any client) . But it is removed in Dynamo Librarie Layout spec. 

@alfarok @ColinDayOrg

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -507,9 +507,10 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
     constructFromIncludeInfo(sortedTypeListNodes, sortedIncludeItemPairs);
 
     // Misc section will take all unprocessed nodes
-    let miscSection = sections.find(section => section.text == miscSectionStr);
-    let unprocessedNodes = sortedTypeListNodes.filter(node => !node.processed);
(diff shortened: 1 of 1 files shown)


src/LibraryUtilities.ts
```diff
@@ -507,9 +507,10 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
     constructFromIncludeInfo(sortedTypeListNodes, sortedIncludeItemPairs);
 
     // Misc section will take all unprocessed nodes
-    let miscSection = sections.find(section => section.text == miscSectionStr);
-    let unprocessedNodes = sortedTypeListNodes.filter(node => !node.processed);
-    unprocessedNodes.forEach(node => buildLibraryItemsFromName(node, miscSection));
+    //https://jira.autodesk.com/browse/QNTM-2976 - Remove the Miscellaneous section
```
ColinDayOrg: Is there a reason for leaving this code here in the commented out state instead of just deleting it?

ColinDayOrg:
LGTM

PR #116, opened 2018-02-23
Send the category data to dynamo

ramramps:
In Dynamo Instrumentation, the Filter categories are collected in 1.3 version of Library.
In the new librarieJS this information is embedded inside the JS. This PR is to add an event which sends this Filter category information to the client (Dynamo in this case).

More specific detail:
Dynamo 1.3 has this for instrumentation
```
  StringBuilder strBuilder = new StringBuilder();
            foreach (var category in SearchCategories)
            {
                strBuilder.Append(category.Name);
                strBuilder.Append(" : ");
                if (category.IsSelected)
                {
                    strBuilder.Append("Selected");
                }
                else
                {
                    strBuilder.Append("Unselected");
                }
                strBuilder.Append(", ");
            }

            Analytics.LogPiiInfo("Filter-categories", strBuilder.ToString().Trim());
```
 The above said `category` is in LibrareJS. It is the `CategoryInfo` class. So, returning that `CategoryInfo` to Dynamo, and it is easy to construct the `data:checked, data:unchecked` for instrumentation

@mjkkirschner @alfarok

### DIFF
--- a/__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
+++ b/__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
@@ -112,6 +112,7 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
       "RefreshLibraryViewRequestName": "refreshLibraryView",
       "SearchTextUpdatedEventName": "searchTextUpdated",
       "SectionIconClickedEventName": "sectionIconClicked",
+      "FilterCategoryEventName": "filterCategoryChange",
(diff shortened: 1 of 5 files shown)


mjkkirschner:
does this require updating the tests?

ramramps:
what tests? there is no tests associated for this, because this is related to UI events.

mjkkirschner:
okay, thought it might matter for the enzyme tests - LGTM

ramramps:
yeah I had the same thought first. but there are no tests for this. But the snapshot is updated. so the event cannot be missed.

PR #117, opened 2018-03-06
Prevent special chars from being executed in search

alfarok:
https://jira.autodesk.com/browse/QNTM-3140

Prevent special chars from being executed in search.  Previously attempting to add a special character would cause library to freeze and would require a complete Dynamo reboot.

### Reviewers
@QilongTang


QilongTang:
LGTM

PR #118, opened 2018-03-22
QNTM-3710: Up and Down arrow keys are not working in library

ColinDayOrg:
This PR is to address the issue of keys not being properly handled in the library when run inside of sandbox or D4R. The keys are properly handled in the librarie.js sample.

The main issue is that the switching on the key code currently is based on the event.code parameter. In the sample file this works as expected. However, in sandbox and D4R the event.code parameter is empty. There is another parameter, event.key, that holds essentially the same information and is valid for all environments. Switching to use this parameter instead causes all environments (sample, sandbox, D4R) to function as expected.

There are three places in librarie.js that currently use event.code;
- LibraryContainer: used for the arrow up and arrow down keys
- Search Bar: used for the escape key
- SearchResultItem: used for the enter key

This PR addresses all three of these areas, with the exception that the escape key is sometimes being eaten by the WPF parts of the sandbox application, and probably the D4R application as well. A separate issue will be filed for that.

All tests have passed, with the exception of two tests that are currently failing in librarie.js due to snapshot issues (see QNTM-3768). These tests fail without these changes.

@mjkkirschner 
@ramramps 
@alfarok

### DIFF
--- a/src/components/LibraryContainer.tsx
+++ b/src/components/LibraryContainer.tsx
@@ -88,13 +88,11 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
     }
 
     handleKeyDown(event: any) {
-        switch (event.code) {
+        switch (event.key) {
             case "ArrowUp":
-                event.preventDefault(); // Prevent arrow key from navigating around search input
                 this.updateSelectionIndex(false);
                 break;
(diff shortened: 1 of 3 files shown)


mjkkirschner:
LGTM

PR #113, opened 2018-01-26
LibrarieJS Updates

ramramps:
This PR 

1. Revert the changes for auto-expand on category.
2. Add-ons are not clickable. Auto expanded by default.
3. Add-Ons icon change 

Note: I did not remove the other changes, but commented out.

screen shot 2018-01-26 at 9 30 15 am

@QilongTang

PR #119, opened 2018-04-19
Enable special characters and encode search string

alfarok:
[QNTM-4042](https://jira.autodesk.com/browse/QNTM-4042)

Supplemental [Dynamo PR](https://github.com/DynamoDS/Dynamo/pull/8776)

This PR enables all characters and spaces in search and encodes the string passed from librarie.js to Dynamo where the string is decoded and run through search.  Previously, spaces were replaced in searches because the results were not accurately returned due to the string formatting.  In a later PR all characters that were not `a-z` `0-9` were removed from the query string to prevent special characters from crashing librarie.  This PR should evade those issues.  This PR also fixes test failures that were introduced prior.

Looking forward we may want to consider modifying the search timeout logic to improve response fluidity.

## Reviewers
@mjkkirschner 

## FYI
@Racel @jnealb @smangarole

### DIFF
--- a/__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
+++ b/__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
@@ -42,6 +42,16 @@ Array [
 >
     String
 </div>,
+  <div
+    className="LibraryItemText"
+>
+    Clockwork
+</div>,
+  <div
+    className="LibraryItemText"
+>
+    DynamoText
+</div>,
 ]
 `;
 
@@ -104,6 +114,7 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
   libraryContainer={
     LibraryController {
       "DefaultSectionName": "default",
+      "FilterCategoryEventName": "filterCategoryChange",
       "ItemClickedEventName": "itemClicked",
       "ItemMouseEnterEventName": "itemMouseEnter",
       "ItemMouseLeaveEventName": "itemMouseLeave",
@@ -112,7 +123,6 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
       "RefreshLibraryViewRequestName": "refreshLibraryView",
       "SearchTextUpdatedEventName": "searchTextUpdated",
       "SectionIconClickedEventName": "sectionIconClicked",
-      "FilterCategoryEventName": "filterCategoryChange",
       "createLibraryByElementId": [Function],
       "createLibraryContainer": [Function],
       "on": [Function],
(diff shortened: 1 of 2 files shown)


mjkkirschner:
@alfarok have you run the tests and made sure all are passing?

alfarok:
@mjkkirschner some tests were actually failing prior to this PR... so those same tests are still failing but no new failures.  Either way, looking into what's going on in there now.

__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
```diff
@@ -41,6 +41,16 @@ Array [
...
+</div>,
+  <div
+    className="LibraryItemText"
+>
+    Clockwork
+</div>,
+  <div
+    className="LibraryItemText"
+>
+    DynamoText
```
alfarok: @mjkkirschner These dependencies are referenced in the `RawTypeData.json` file and will cause the comparison to fail if they are not included here.  I am not 100% what changed recently that would make this a new requirement, @ramramps do you have any ideas?  As far as I can tell these packages have been in the `RawTypeData.json` file for a while.

__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
```diff
@@ -112,7 +123,6 @@ exports[`LibraryContainer Test UI rendering of Library Item and child components
       "RefreshLibraryViewRequestName": "refreshLibraryView",
       "SearchTextUpdatedEventName": "searchTextUpdated",
       "SectionIconClickedEventName": "sectionIconClicked",
-      "FilterCategoryEventName": "filterCategoryChange",
```
alfarok: @ramramps I noticed this was also a recent addition in #116.  Tests also fail for me if these are not in this order, appears to be alphabetical

alfarok: No it needs to be present but the order in which the objects are listed appears to influence the test results.

__tests__/Snapshottest/__snapshots__/UIOutputComparisonTests.tsx.snap
alfarok: Everything passes if I include `Clockwork` and `DynamoText` in the snap.  I wonder if changes with any of the element types made this a requirement at some point.  It's hard to tell exactly when these tests broke without pulling old code and running them until everything passes

alfarok: I think they first start failing at #113 where tests were added around Add-ons

ramramps: ah ok..

PR #120, opened 2018-07-02
Updates for AVP

ramramps:
This PR is to update LibrarieJS to support AVP. 

The changes are very minimal, and it should not affect Dynamo.

@mjkkirschner @alfarok @QilongTang

### DIFF
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -129,7 +129,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
             if (isThereChildItemsToExpand.length == 0) {
                 setTimeout(() => {
                     let elem = ReactDOM.findDOMNode(this);
-                    elem.scrollIntoView();
+                    elem.scrollIntoView(false);
                 }, 0);
             }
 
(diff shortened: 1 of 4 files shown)


src/components/LibraryItem.tsx
```diff
@@ -460,7 +460,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         if (this.props.data.childItems.length == 0) {
             let rec = ReactDOM.findDOMNode(this).getBoundingClientRect();
             let mouseEnterEvent = libraryContainer.props.libraryController.ItemMouseEnterEventName;
-            libraryContainer.raiseEvent(mouseEnterEvent, { data: this.props.data.contextData, rect: rec });
+            libraryContainer.raiseEvent(mouseEnterEvent, { data: this.props.data.contextData, rect: rec, element:  ReactDOM.findDOMNode(this) });
```
ramramps: should not affect dynamo here. the element is sent  here, so avp can add tooltip on top of it

src/components/SearchBar.tsx
```diff
@@ -95,7 +95,9 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
                 this.clearInput();
                 break;
             default:
-                this.searchInputField.focus();
+                if (event.target.className == "SearchInputText") {
+                    this.searchInputField.focus();
```
ramramps: the key handlers are global. codeblock has a text field, any keystroke will get the search in focus. this change should get rid of that

PR #121, opened 2018-07-20
search improvments

mjkkirschner:
https://jira.autodesk.com/browse/QNTM-4324

This PR does the following:

* no longer url encodes the the search text, it's not necessary. For clients who need to do this before sending the results to their own search algorithm - it should be done in client code - See related PR in Dynamo.

* makes the `getHighLightText`  method more robust by escaping regex meta characters, and wrapping the regex matching in a try catch - if highlight text is not found or an exception is thrown, we just return an empty array of spans to highlight, and search continues to work - without the text highlight feature.

* adds tests for searching for items with special characters like " ", "." and "$"

I will send 2 more PRS - one to Dynamo and one to AVP with updated version of librariejs built from this branch.

### Reviewers:
@ramramps  @alfarok


mjkkirschner:
@ramramps can I merge this?

ramramps:
LGTM

PR #127, opened 2019-09-17
Delete duplicate folder

jesusHCG:
### Description
Solve a problem where compilation cannot run in a Linux environment

### Related GitHub issue or JIRA story
https://jira.autodesk.com/browse/DYN-2071
https://jira.autodesk.com/browse/DYN-2072

### Checklist
- [ ] PR title reflects the changes it introduces. Consider using this template: QNTM-XXXX: Implement X Feature
- [ - ] Code compiles correctly
- [ - ] No linting errors
- [ - ] All tests passing
- [ ] Updated README.md / documentation (if necessary)


mjkkirschner:
@QilongTang this looks like the right solution, but we may want to see if it causes any issues with Dynamo build - I am not sure if we copy these files to Dynamo when we cross merge - it likely does not matter so much right now as Dynamo is only being consumed on windows platforms.

mjkkirschner:
LGTM

mjkkirschner:
https://github.com/DynamoDS/Dynamo/tree/master/src/LibraryViewExtension/web
this is what we have in Dynamo - I think copying the min file is usually all we need to do - so I don't anticipate any other changes - what do you think @jesusHCG ?

jesusHCG:
These were the specific files that were move from the folder.
image

I don't think it could be a reason for an error, but we can perform some tests when making the min file

QilongTang:
@jesusHCG @mjkkirschner I think this change should be safe. LGTM. These svgs are duplicated in Dynamo repo so should not impact Dynamo at all.

PR #130, opened 2019-10-03
Add of csproj file for nuget packing

jesusHCG:
### Description
Change of Artifactory url and docker image for the custom whit .net core, and add the custom deploy options to pipeline.yml

### Related GitHub issue or JIRA story
https://jira.autodesk.com/browse/DYN-2117
https://jira.autodesk.com/browse/DYN-2174
https://jira.autodesk.com/browse/DYN-2175

### Checklist
- [ ] PR title reflects the changes it introduces. Consider using this template: QNTM-XXXX: Implement X Feature
- [ - ] Code compiles correctly
- [ - ] No linting errors
- [ - ] All tests passing
- [ ] Updated README.md / documentation (if necessary)


mjkkirschner:
@jesusHCG what actual files make it into the nupkg file - can you add a screenshot?

jesusHCG:
Only librarie.min.js and the resources

image

librariejs.csproj
```diff
@@ -0,0 +1,31 @@
...
+  <PropertyGroup>
+    <Description>Project for nuget pack</Description>
+    <Authors>Autodesk</Authors>
+    <PackageTags>Dynamo;librariejs</PackageTags>
+    <NeutralLanguage>en-US</NeutralLanguage>
+    <versionprefix>1.0.0</versionprefix>
+    <PackProjectInputFile>$(MSBuildProjectFullPath)</PackProjectInputFile>
+    <NoBuild>true</NoBuild>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
```
mjkkirschner: can you change this to .net 4.7 - as that is how we will the project version we will consume this from for now.

jesusHCG: We can´t, because the docker doesn't have the full net standard framework and causes conflicts on the dotnet packing.

mjkkirschner: okay, we need to make sure that we can actually reference this in Dynamo then. Theres no point in publishing it if Dynamo.sln cannot consume it.

Alternatively instead of using a csproj - you can just use nuget and .nuspec file like this:
https://github.com/DynamoDS/Dynamo/blob/master/tools/NuGet/template-nuget/DynamoVisualProgramming.Core.nuspec

and call nuget directly to package the nuspec into a nupkg

jesusHCG: This is the error that we got, whit 4.7 version.

image

jesusHCG: We can´t use nuget commands on a docker linux container. We also tried the .nuspec file on several tests.

mjkkirschner: thanks @jesusHCG we might need to go back to the drawing board on this then - can you send me the generated nupkg file - we need to be certain Dynamo.sln can actually consume it - we should not require devs to install dotnetcore 2 just to consume this single package.

If it does not work we'll need to figure something else out like using a windows docker instance.

mjkkirschner: What about using: 
--no-build option when doing the -pack

alfredo-pozo: @mjkkirschner The nupkg will work for Dynamo, since we are going to do similar steps as we are doing with the ASM lib, install the package on an specific location and copy the necessary files to the project (min.js and resources)

mjkkirschner: I don't think that is the ideal workflow - I would ideally like the nuget reference added to the librarieJSExtension.csproj file - this way we can easily use it or update it as part of the dynamo.sln in any build job.

mjkkirschner: consider the case that dynamo repo is being built in another build pipeline, github actions, etc or on a users machine (it is open source) etc - 
it's easiest and *best* if we can take advantage of nuget restore within visual studio, not need to update some config file somewhere to change the reference to a new version of librariejs.

alfredo-pozo: @mjkkirschner I have a question there, if the plan is to keep it up inside .sln, so anyone can build and take advantage of the restore, isn't going to generate an error since the .nupkg will be available only on ADSK artifactory? Or are we going to publish it to nuget.org?

mjkkirschner: published to nuget - dynamo always needs to build for people outside ADSK.

pipeline.yml
```diff
@@ -14,18 +14,27 @@ nodejs_version: "10"
...
+
+deployment:
+-
+  type: customized
+  customized_builder_image:
+    registry: artifactory
+    url: artifactory.dev.adskengineer.net/docker-local-v2/dynamo/dynamolibrariejs:dotnetcore2
+  scripts:
+    - "dotnet pack -c Release -o nuget --version-suffix $BUILD_NUMBER"
+    - "dotnet nuget push \"$(find nuget -type f -name *.nupkg)\" -s https://art-bobcat.autodesk.com/artifactory/api/nuget/team-dynamo-nuget -k $API_KEY"
```
mjkkirschner: did you verify with @smangarole that we want to publish this to internal nuget?
I cannot recall 100% but I thought we discussed publishing to nuget as this is an open source repo that Dynamo will know rely on.

librariejs.csproj
alfredo-pozo: Ok, since the ideal workflow is totally different from the one that we agree before on the standups, we need to redirect the efforts to a different approach/solution

pipeline.yml
alfredo-pozo: @mjkkirschner  Please take a look on that, we always talk about the artifactory as the ADSK aritifactory: https://jira.autodesk.com/browse/DYN-2117

librariejs.csproj
```diff
@@ -1,7 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Description>Project for nuget pack</Description>
+    <Description>Reusable lightweight library component based on React.js</Description>
     <Authors>Autodesk</Authors>
+    <Copyright>Autodesk 2019</Copyright>
+    <PackageProjectUrl>https://github.com/DynamoDS/librarie.js</PackageProjectUrl>
```
mjkkirschner: @jesusHCG this should be MIT license - see the repo LICENSE file.

jesusHCG: it´s already added.

PR #133, opened 2020-01-02
Trident web browser control compat

mjkkirschner:
@reddyashish @QilongTang 
Here are the changes I had to make to librarieJS to support this PR:

https://github.com/DynamoDS/Dynamo/pull/10241

I prefer to keep this a separate branch for now, wanted to generate the diff for review.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://art-bobcat.autodesk.com/artifactory/api/npm/autodesk-npm-virtual/
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -179,6 +179,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     }
 
     onImageLoadFail(event: any) {
+        event.target.orgSrc = event.target.src;
         event.target.src = require("../resources/icons/default-icon.svg");
     }
 
--- a/src/entry-point.tsx
+++ b/src/entry-point.tsx
@@ -38,7 +38,6 @@ interface SearchLibraryItemsFunc {
 export class LibraryController {
 
     ItemClickedEventName = "itemClicked";
-    AnyItemClickedEventName = "anyitemClicked";
     ItemMouseEnterEventName = "itemMouseEnter";
     ItemMouseLeaveEventName = "itemMouseLeave";
     ItemSummaryExpandedEventName = "itemSummaryExpanded";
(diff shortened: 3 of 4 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://art-bobcat.autodesk.com/artifactory/api/npm/autodesk-npm-virtual/
```
QilongTang: Didn't we make this change before already?

src/entry-point.tsx
```diff
@@ -4,6 +4,8 @@ import * as ReactDOM from "react-dom";
 import { LibraryContainer } from "./components/LibraryContainer";
 import { JsonDownloader } from "./LibraryUtilities";
 import { Reactor, Event } from "./EventHandler";
+import "core-js";
```
QilongTang: Hi @mjkkirschner I did not understand the purpose from your commit message, can you elaborate?

.npmrc
mjkkirschner: yes, but only on the jenkins branch.... which I think we should get rid of - all development should be on master to avoid things getting out of sync.

mjkkirschner: https://github.com/DynamoDS/librarie.js/tree/develop

mjkkirschner: it already is :( out of sync.

src/entry-point.tsx
mjkkirschner: core-js includes a bunch of polyfills - so that browsers (like ie) which are missing some functions can use them - in this case I needed `string.prototype.startsWith` but we could probably just add the polyfill for that single method.... this just works though.

src/components/LibraryItem.tsx
```diff
@@ -179,6 +179,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
     }
 
     onImageLoadFail(event: any) {
+        event.target.orgSrc = event.target.src;
```
mjkkirschner: this lets us lookup what the originalSrc url was for an icon, even if we have fallenback to the default dynamo icon.

.npmrc
QilongTang: @mjkkirschner Thx for the reminder!

mjkkirschner:
@QilongTang @zeusongit @reddyashish do you guys have a preference to between merging this into master or to keeping it on a separate branch?

aparajit-pratap:
I am for having to maintain one version of librarie.js as far as possible so yes, I would go for `master`.

QilongTang:
`master` is fine with me

mjkkirschner:
tests are all passing on this branch. I am going to file a task to delete the develop branch and merge it to master.

PR #167, opened 2021-11-25
update styling

alvpickmans:
Updating font sizes to latest design and reducing negative space between search bar and library items.

It seems to be an issue with certain machines using a different IE engine for WPF's WebBrowser component in Dynamo that don't support certain css rules (ie `padding: 1rem 2rem`), so splitting most of them using `rem` units into each `padding-{SIDE}` rule.

library-js-update

### DIFF
--- a/src/resources/LibraryStyles.css
+++ b/src/resources/LibraryStyles.css
@@ -26,6 +26,10 @@ input {
     font-family: "Artifakt Element", "Open Sans";
 }
 
+button{
+    font-size: 1rem;
+}
+
 .LibraryContainer {
     max-height: 100vh;
     display: flex;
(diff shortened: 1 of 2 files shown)


alvpickmans:
cc @SHKnudsen @QilongTang

src/resources/LibraryStyles.css
```diff
@@ -393,6 +428,10 @@ input {
     cursor: pointer;
 }
 
+.SearchBar button.CancelButton:hover{
```
zeusongit, line 431: This is also defined at line `659`, could be consolidated into a single definition. Nothing big though.

alvpickmans: this has been unified under the same block now, thanks!

reddyashish (approved):
Changes look good to me. Thank you @alvpickmans, I see this is also fixing some of the improvements from the https://jira.autodesk.com/browse/DYN-4396. Are you planning on pushing any more commits to this PR?

reddyashish:
Merging this. Dynamo repo PR can be merged once the checks are finished.

PR #171, opened 2022-02-09
Fix library search and display add-ons section always.

reddyashish:
Fixes the library search and display add-ons section always.

Library

@QilongTang @mjkkirschner

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -563,11 +563,16 @@ export function removeEmptyNodes(items: ItemData[]) {
             if (removeEmptyNodes(item.childItems)) {
                 i--;
             }
-        }else if ( (item.itemType === "section" && item.contextData !== "Add-ons") || item.itemType === "category" || item.itemType === "group" || item.itemType === "coregroup") {
+        }
+		// Do not remove the Add-ons section from the list even when it has no child elements, as we want to show this section at all times.
+		else if (item.itemType === "section" && item.contextData === "Add-ons") {
+			continue;
+		}
+		else if ( item.itemType === "section" || item.itemType === "category" || item.itemType === "group" || item.itemType === "coregroup") {
             items.splice(i, 1);
             i--;
             itemRemoved = true;
-        }// Do not remove the Add-ons section even when it has no child elements, as we want to show this section at all times.
+        }
     }
 
     return itemRemoved;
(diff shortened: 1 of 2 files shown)


src/components/SearchResultItem.tsx
```diff
@@ -81,46 +81,55 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
         let ItemContainerStyle = this.state.selected ? "SearchResultItemContainerSelected" : "SearchResultItemContainer";
         let iconPath = this.props.data.iconUrl;
 
-        // The parent of a search result item is the second last entry in 'pathToItem'
-        let parentText = this.props.data.pathToItem[this.props.data.pathToItem.length - 2].text;
```
QilongTang, line 85: Oh so this is returning null in certain cases?

reddyashish: yeah, this.props.data.pathToItem.length was less than 2 in case of sections. So that was being null.

mjkkirschner: yeah... maybe a comment is in order, what is this checking for?

src/LibraryUtilities.ts
```diff
@@ -563,11 +563,16 @@ export function removeEmptyNodes(items: ItemData[]) {
             if (removeEmptyNodes(item.childItems)) {
                 i--;
             }
-        }else if ( (item.itemType === "section" && item.contextData !== "Add-ons") || item.itemType === "category" || item.itemType === "group" || item.itemType === "coregroup") {
+        }
+		// Do not remove the Add-ons section from the list even when it has no child elements, as we want to show this section at all times.
+		else if (item.itemType === "section" && item.contextData === "Add-ons") {
```
mjkkirschner, line 568: how is this different from your original changes?

src/components/SearchResultItem.tsx
reddyashish: Added a comments to that and we have a test for it which was added in the last PR https://github.com/DynamoDS/librarie.js/blob/master/__tests__/mochatest/libraryUtilitiesTest.ts#L119

src/LibraryUtilities.ts
reddyashish: These changes are same, just added it to a different else block to be clear. As we are not removing the add-ons section here, had to do the filtering on the search side.

PR #174, opened 2022-03-23
Secissues march2022 - webpack 4

mjkkirschner:
same intention as:
https://github.com/DynamoDS/librarie.js/pull/173

but we only update to webpack 4 and its peer deps instead of webpack 5.

⚠️ After this there are two remaining issues in the harmony list, and 10 vulns reported by npm audit that cannot be fixed without using incompatible versions.

Both harmony vulns are in deps of webpack itself, so I think we are safe to mark these dev deps.

big changes here:
* webpack 4
* adds webpack cli peer dep
* updates webpack loader peer deps
* adds webpack analyzer
* updates webpack config to target web instead of node.
* updates typescript
* updates tests for new typescript rules
* swaps from awesome-typescript-loader to ts-loader - (not maintained)
* removes jest-jenkins-reporter and swaps it for jest-junit, source of a bunch of issues. (not maintained)
* removes some packages we do not reference anymore - appears they were added to try to force certain versions, but this doesn't work because npm will pull nested node_modules folders in some cases.
* update core-js
* includes only some of the core-js, instead of everything
* drops font-awesome from repo and css - I did not see it used after visual refresh and it was big.
* move from uglify to terser for minification - part of webpack4 update. (builtin)
* set mode based on prod or dev.
* fix busted .svg files. (+ symbol)
* 

tested in dynamo, performance looks good.


pinzart90 (approved):
LGTM

webpack.config.js
```diff
@@ -1,40 +1,36 @@
 "use strict";
 const webpack = require('webpack');
+const TerserPlugin = require("terser-webpack-plugin");
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
sm6srw, line 4: Should this also be commented?

sm6srw (approved):
One question then LGTM!

webpack.config.js
mjkkirschner: nah, this doesn't harm anything to load at build time, it does not get bundled in any case.

PR #176, opened 2022-04-05
move to node 16

mjkkirschner:
updates docker build image to node 16
harmony still times out, but will send a separate PR for that.


sm6srw (approved):
LGTM!

PR #178, opened 2022-08-24
Use adsk-npm-license-puller (internal tool) to generate about box content

mjkkirschner:
This PR does the following:
* adds a dev dependency on `adsk-npm-license-puller` of the latest version (thanks @twastvedt for your work here)
* this tool will only exist when using adsk npm registry!
* adds a npm script `generate_license` which will create two different about box .html files.

Next:
- [x] send a similar PR to notification center
- [ ] in Dynamo - consume the about box html with a webBrowser control (or similar) in the about box.

### DIFF
--- a/license_output/about-box_direct.html
+++ b/license_output/about-box_direct.html
@@ -1,0 +1,141 @@
+<div>
+  <p>
+    <strong>
+        'librarieJS' &copy; 2024 Autodesk, Inc. All rights reserved.
+    </strong>
+  </p>
+  <div>
+    <p>
+      <strong>
+            All use of this Software is subject to the Autodesk terms of service accepted upon access or use of this Service or made available on the Autodesk webpage. Autodesk terms of service for Autodesk&rsquo;s various web services can be found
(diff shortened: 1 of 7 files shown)


QilongTang:
@mjkkirschner Is the duplication intended?

mjkkirschner:
@QilongTang yes, the script I added generates both the direct dependencies and transitive dependencies about boxes for ease of tracking and integration., and because it's not clear what is the best practice - for now I am planning to integrate the direct dependency about box content, but having the full list in the git history is very useful over time I think.

If possible it would be nice to ignore this folder from the sonar scan - do you know if this is possible?

pinzart90 (approved):
LGTM

license_output/about-box_direct.html
```diff
@@ -0,0 +1,141 @@
+<div>
```
pinzart90, line 1: not sure where you got this file from...

mjkkirschner: the npm puller generates these files in the license_output folder.

PR #180, opened 2023-01-20
Fix harmony scans

pinzart90:


pipeline.yml
```diff
@@ -56,12 +56,11 @@ deployment:
 soc2:
   run_on_any_branch : true
   harmony:
+    product_name: Dynamo
     hidden_email_list: dynamo_ws_access
     repository: env.HARMONY_REPO
-    third_party_lib_paths:
```
pinzart90, line 61: harmony knows to look for package.json and node_modules in js projects

mjkkirschner (approved):

aparajit-pratap (approved):
So was the product name missing, which was the cause for the harmony scans not running?

pinzart90:
> So was the product name missing, which was the cause for the harmony scans not running?

The product name was the default, git org name. This did not correspond to the security dashboard

PR #186, opened 2023-08-04
hotfix: remove onTextChanged execution within forward delete

Enzo707:
This PR removes onTextChanged execution within the forward delete implementation related to [DYN-6064](https://jira.autodesk.com/browse/DYN-6064). This onTextChanged is causing some unexpected behaviour while expanding nodes 

**FYI** 
@RobertGlobant20 
@QilongTang

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -84,7 +84,6 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
     }
 
     UNSAFE_componentWillMount() {
-        // window.handleKeyDown = this.handleKeyDown();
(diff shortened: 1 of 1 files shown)


src/components/SearchBar.tsx
```diff
@@ -140,7 +139,6 @@ export class SearchBar extends React.Component<SearchBarProps, SearchBarState> {
         this.searchInputField.value = searchValueCopy.join("");
         this.searchInputField.focus();
         this.searchInputField.setSelectionRange(cursor, cursor);
-        this.props.onTextChanged(this.searchInputField.value);
```
QilongTang, line 143: But wouldn't you want to update the value (search term) after the forward delete action?

PR #188, opened 2023-08-15
remove(LibraryItem):forced scroll at componentDidMount

Enzo707:
This PR removes an useless block of code that causing an unexpected scroll behaviour in library

### DIFF
--- a/src/components/LibraryItem.tsx
+++ b/src/components/LibraryItem.tsx
@@ -117,26 +117,6 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
         // }
     }
 
-    //Afer rendering each Library item, scroll to the expanded item
-    componentDidMount() {
-        if (this.props.data.expanded && this.props.data.itemType !== "coregroup") {
-            //scroll to only that element clicked from search. Determining that element from 
-            //other elements is little tricky. The idea here is, the element which has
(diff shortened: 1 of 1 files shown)


src/components/LibraryItem.tsx
```diff
@@ -117,26 +117,6 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
...
     }
 
-    //Afer rendering each Library item, scroll to the expanded item
-    componentDidMount() {
-        if (this.props.data.expanded && this.props.data.itemType !== "coregroup") {
-            //scroll to only that element clicked from search. Determining that element from 
-            //other elements is little tricky. The idea here is, the element which has
-            //its child elements expanded to false is the actual element clicked from search. Scroll
-            //to that element.
-            let isThereChildItemsToExpand = this.props.data.childItems.filter((item: any) => {
```
QilongTang, line 127: @Enzo707 I remember this function is used so that clicking from library search results would actually scroll to the item clicked, after removing this function, what would be the behavior in that case?

Enzo707: @QilongTang That's what the comments say but as far as I can tell this was used just for make the add-ons always visible. If we want so, we need to choose a different approach instead in order to avoid this force scroll.

QilongTang (approved):

PR #189, opened 2023-08-18
refactor/DYN-6124/code-smells

Enzo707:
after-code-smells-fixes-min (3)

This PR includes refactoring due to code smells reports.

- Removes useless functions and commented legacy code
- Updates according to ES6+ and javaScript recommended standards

**FYI**
@QilongTang 
@RobertGlobant20

### DIFF
--- a/src/LibraryUtilities.ts
+++ b/src/LibraryUtilities.ts
@@ -6,8 +6,6 @@ type MemberType = "none" | "create" | "action" | "query";
 type ElementType = "none" | "section" | "category" | "group" | "coregroup" | "classType";
 type ItemType = "none" | "section" | "category" | "group" | "create" | "action" | "query" | "classType" | "coregroup";
 
-import * as _ from 'underscore';
-
 export class TypeListNode {
 
     fullyQualifiedName: string = "";
@@ -64,12 +62,12 @@ export class LayoutElement {
         this.include = data.include;
 
         if (data.childElements) {
-            for (let i = 0; i < data.childElements.length; i++) {
-                this.childElements.push(new LayoutElement(data.childElements[i]));
+            for (const element of data.childElements) {
+                this.childElements.push(new LayoutElement(element));
             }
         }
 
-        if (data.showHeader == false) {
+        if (data.showHeader === false) {
             this.showHeader = data.hideSectionHeader;
         }
     }
@@ -116,7 +114,6 @@ export class ItemData {
         }
 
         // Get keywords from typeListNode and push them into itemData
-        let keywords = typeListNode.keywords.split(",");
         this.keywords.forEach(keyword => {
             this.keywords.push(keyword.toLowerCase().replace(/ /g, ''));
         });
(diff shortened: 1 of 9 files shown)


src/LibraryUtilities.ts
```diff
@@ -492,13 +489,10 @@ export function buildLibrarySectionsFromLayoutSpecs(loadedTypes: any, layoutSpec
         sectionElements.push(new LayoutElement(section));
     }
 
-    //call update category groups.
-    updateCategoryGroups(sectionElements);
```
RobertGlobant20, line 496: is this call not necesary? or is done in other place?

src/LibraryUtilities.ts
```diff
@@ -825,7 +784,7 @@ export function findAndExpandItemByPath(pathToItem: ItemData[], allItems: ItemDa
     );
 
     if (pathToItem.length == 1) {
-        return item ? true : false;
+        return !!item;
```
RobertGlobant20, line 787: could you explain why` item ? true : false; `is the same than` !!item;`?
I know that "!" is the logical not operator but not sure why is used two times?

RobertGlobant20 (commented):
LGTM with two comments.
Also did you test this changes in DynamoSandbox? Please include a GIF of your testing.

src/LibraryUtilities.ts
Enzo707: @RobertGlobant20 Please, let me know if I'm missing something but If you look at the function declaration, it doesn't return anything that might affect the context where it is executed. Also, all the operations there are related to `sectionElements` which is defined as a `let`.

src/LibraryUtilities.ts
Enzo707: @RobertGlobant20 !! operator returns the boolean value. [Ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!)

src/LibraryUtilities.ts
```diff
@@ -116,7 +114,6 @@ export class ItemData {
         }
 
         // Get keywords from typeListNode and push them into itemData
-        let keywords = typeListNode.keywords.split(",");
```
QilongTang, line 119: why do we remove this line?

Enzo707: @QilongTang This variable is not used. The operations below this statement refers to keywords property of class ItemData class (`this.keywords`)

src/components/LibraryItem.tsx
```diff
@@ -45,31 +44,15 @@ class GroupedItems {
...
-                case "query": this.queries.push(items[i]); break;
-                default: this.others.push(items[i]); break;
+            switch (element.itemType) {
+                case "create": this.creates.push(element); break;
+                case "action": this.actions.push(element); break;
+                case "query": this.queries.push(element); break;
+                default: this.others.push(element); break;
             }
 
-            var finalData = {
```
QilongTang, line 57: I guess this is also not used?

src/LibraryUtilities.ts
QilongTang: yeah looks like the function was disabled so no longer valid

QilongTang (commented):
LGTM with one comment

src/components/LibraryItem.tsx
Enzo707: Right, it is not used.

QilongTang (approved):

QilongTang:
Please make the change to update librarie.js on Dynamo `master` branch only

PR #190, opened 2023-09-06
[DYN-6217]update: package/package-lock after removing node-notifier dependency

Enzo707:
This PR updates the package and package-lock file after removing node-notifier dependency in order to avoid security issues.

**FYI**
@QilongTang


aparajit-pratap:
I would defer this review to @mjkkirschner as he said he added this as a peer dependency. Does that mean the `jest` package requires this specific version of `node-notifier`? If so, then we might have to update `jest` so it doesn't require this older version of node-notifier, wouldn't we?

Enzo707:
> I would defer this review to @mjkkirschner as he said he added this as a peer dependency. Does that mean the `jest` package requires this specific version of `node-notifier`? If so, then we might have to update `jest` so it doesn't require this older version of node-notifier, wouldn't we?

I thought @mjkkirschner say that if we remove node-notifier as direct dependency, the tests pass and the build goes ok that should be it. Please confirm

QilongTang:
Thank for confirming @aparajit-pratap @mjkkirschner

PR #194, opened 2023-10-23
DYN-6270/feature/refactor based on sonarcloud reports

Enzo707:
This PR includes refactoring based on sonarcloud suggestions

**Review**
@reddyashish 
@QilongTang

### DIFF
--- a/src/resources/LibraryStyles.css
+++ b/src/resources/LibraryStyles.css
@@ -14,7 +14,7 @@ html {
 }
 
 body {
-	font-family: "Artifakt Element", "Open Sans";
+	font-family: "Artifakt Element", "Open Sans", sans-serif;
 	-webkit-user-select: none;
 	user-select: none;
 	cursor: default;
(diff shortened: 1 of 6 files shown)


src/resources/LibraryStyles.css
```diff
@@ -429,27 +429,12 @@ button {
...
 	background-color: transparent;
 	transition: 0.15s;
-}
-
-.SearchBar button {
-	color: #ddd;
-	transition: 0.15s;
-}
-
-.SearchBar button {
```
QilongTang, line 446: this is not used?

Enzo707: @QilongTang the `.SearchBar button` selector was repeated several times so I combined all styles applied. Those I've left there are the used ones

RobertGlobant20: by checking the code I can see that the value
`color: #ddd;`
is set several times to different values, then if you combined all the styles for this button, how did you know which is the right color?

Enzo707: @RobertGlobant20  If load the component using the browser, opening devtools you will see the styles which are actually in use.
Image 30-10-23 at 12 50

QilongTang (approved):

PR #196, opened 2023-11-01
DYN 6270 library cleanup based on sonarcloud reports

Enzo707:
This PR includes changes that improves the quality of the code based on sonarcloud reports. 
Most of the changes are regarding to types issues, useless fragments of code, unused imports or variables and var, let, const correct usage.

**Review**
@QilongTang 
@RobertGlobant20

### DIFF
--- a/src/components/SearchResultItem.tsx
+++ b/src/components/SearchResultItem.tsx
@@ -13,7 +13,6 @@ interface SearchResultItemProps {
     libraryContainer: LibraryContainer,
     highlightedText: string;
     detailed: boolean;
-    showItemSummary: boolean;
     onParentTextClicked: ParentTextClickedFunc,
 }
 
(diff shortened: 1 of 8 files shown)


src/components/SearchResultItem.tsx
```diff
@@ -13,7 +13,6 @@ interface SearchResultItemProps {
     libraryContainer: LibraryContainer,
     highlightedText: string;
     detailed: boolean;
-    showItemSummary: boolean;
```
QilongTang, line 16: @Enzo707 Is this property not used?

QilongTang (commented):
LGTM with one comment

src/components/SearchResultItem.tsx
Enzo707: @QilongTang I haven't seen any case of use

PR #198, opened 2023-11-09
DYN-6270 Bugfix webpack.config regex rule for asset resource

Enzo707:
Fix the regex rule applied to asset resources that made library build fail.

**REVIEW**
@avidit


avidit (approved):
LGTM

PR #201, opened 2023-12-06
fix(index.html): restore createLibraryByElementId execution

Enzo707:
The component wasn't rendering properly isolatedly and locally

Before:
image

After:
974CF1D0-2D0A-45FE-96E4-35F6722858FC


avidit (approved):
LGTM

PR #224, opened 2024-04-22
DYN-6739 Regression Typing Fast in Search

RobertGlobant20:
When typing fast in the TextSearch box was getting wrong results or the results were not in the same order than Dynamo InCanvasSearch, 
It was happening due that we are implementing the React Debounce pattern (this patters allows the user to continue typing while we are searching ) but we were missing to update the results immediately after getting the results so I added a call to this.updateSearchViewDelayed() method immediately after getting the results.

FYI: @reddyashish @Enzo707

### DIFF
--- a/src/components/LibraryContainer.tsx
+++ b/src/components/LibraryContainer.tsx
@@ -256,15 +256,14 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
         if (!this.generatedSections) return;
 
         clearTimeout(this.timeout);
-
-        let hasText = text.length > 0;
-
-        if (hasText) {
-            // Starting searching immediately after user input, 
-            // but only show change on ui after 300ms
-            setTimeout(function () {
-                if (this.props.libraryController.searchLibraryItemsHandler) {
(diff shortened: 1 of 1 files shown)


RobertGlobant20:
GIF showing the expected behavior when typing fast.
nZjfIGgNMM

Enzo707:
@RobertGlobant20 there's a couple of side effect I've noticed:
-There's still in search mode after cleaning up the search input
-If you try to perform any change within canvas search the focus seems to be still in library search.

C9Wg2oapWg

RobertGlobant20:
> @RobertGlobant20 there's a couple of side effect I've noticed:
> 
> * There's still in search mode after cleaning up the search input
> * If you try to perform any change within canvas search the focus seems to be still in library search.
> 
@Enzo707 The first bug was fixed in the next commit:
https://github.com/DynamoDS/librarie.js/pull/224/commits/663777109e562603dedca31c03c92a791b0602a7

The second bug is an existing bug you can test it in previous versions of dynamo.

here is the expected behavior.
5iIgpuqx7F

QilongTang (approved):

reddyashish (approved):

PR #228, opened 2024-05-09
DYN-6956 - symbol nodes should not show in home workspace search/browse UI.

mjkkirschner:
This PR adds some new state to the `LibraryContainer` and some conditional rendering to the `LibraryItem` based on props from the `LoadedType` data.

`LibraryContainer.State.HostContext` is essentially an enum that can be used from the Dynamo side to let LibrarieJS know what context it's being displayed in (home workspace, custom node workspace etc)

`LibraryContainer.State.shouldOverideExpandedState ` also controls the rendering behavior of LibraryItems - I found that LibraryItems today use a legacy life cycle hook `UNSAFE_onWillRecieveProps` - to override state with props - specifically for the expanded state so that when interacting only one library item can be open at a time.

This hook also makes it so that when the UI is rendered the currently open item will close. Sometimes this is not desirable like when switching between workspace contexts, so I added some new state which is essentially a mode to control this behavior - currently we turn off the behavior only when switching contexts, so that library items retain their expanded state during the context change render.

LoadedTypes now have `hiddenInWorkspaceContext` which determines if they are hidden in the home workspace context.

I could not figure out how to add a test for this behavior but it seems straight forward for someone familiar with jest/enzyme - I am out of bandwidth and will file a task for it.


src/entry-point.tsx
```diff
@@ -166,4 +171,19 @@ export class LibraryController {
...
+
+    modifyLibraryItemData(context:any ,data:ItemData):boolean{
+        if(this.modifyItemHandler){
+            return this.modifyItemHandler(context,data);
+        }
+       return false;
+    }
+    //TODO summary
+    setHostContext(context:HostingContextType):void{
+        console.log("I AM HERE")
```
RobertGlobant20, line 183: Where are you?

src/components/LibraryContainer.tsx
```diff
@@ -293,6 +303,7 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
 
     updateSearchViewDelayed(text: string) {
         if (text.length == 0) {
+            LibraryUtilities.setItemStateRecursive(this.generatedSections, true, false);
```
RobertGlobant20, line 306: If you search in the code when the updateSearchViewDelayed() is called the LibraryUtilities.setItemStateRecursive() method was already called then why call this method again?

RobertGlobant20 (approved):
LGTM with some comments

src/sharedTypes.ts
```diff
@@ -0,0 +1,3 @@
+//Todo better name
+export type HostingContextType = "home"|"custom"|"none"
```
pinzart90, line 2: are we expecting more types here ?
do these values ("home"|"custom"|"none") correspond to anything in Dynamo ?

src/entry-point.tsx
mjkkirschner: hah, will remove. This reminds me, I had really bad luck trying to debug the jest tests - have you or @EnzoBatistaU been able to get that working?

src/sharedTypes.ts
mjkkirschner: yes, I can make it clearer with a comment.
"home" is a homeworkspace, custom is customNodeWorkspace. I don't think we'd expect anything else here - and currently only "home" does anything.

src/components/LibraryContainer.tsx
mjkkirschner: yeah, I needed to do this when I had reverted that `onWillRecieveProps` hook where the props override state, so expanded was not getting reset - I will see if I can remove this now.

src/sharedTypes.ts
pinzart90: Why not an Enum ?
It would be easier to compare, I think.
HostingContextType.Home == state.hostContext;

src/components/SearchResultItem.tsx
```diff
@@ -46,14 +47,14 @@ export class SearchResultItem extends React.Component<SearchResultItemProps, Sea
         if (this.state.selected) {
             let container = ReactDOM.findDOMNode(this.props.libraryContainer);
             let currentItem = ReactDOM.findDOMNode(this);
-            let containerRect = container.getBoundingClientRect();
-            let currentRect = currentItem.getBoundingClientRect();
+            let containerRect = container?.getBoundingClientRect();
+            let currentRect = currentItem?.getBoundingClientRect();
 
-            if (currentRect.top < currentRect.height) {
+            if (currentRect?.top < currentRect?.height) {
```
pinzart90, line 53: so if these variables are null, what should this operator (<) return ?

src/components/LibraryContainer.tsx
```diff
@@ -352,6 +363,12 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
         }
     }
 
+    setOverride(override:boolean){
+        if(this.state.shouldOverideExpandedState != override){
+        this.setState({shouldOverideExpandedState:override})
+        }
+    }
+
```
pinzart90, line 371: what is this used for ?

src/components/LibraryContainer.tsx
```diff
@@ -374,6 +391,7 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
                         data={data}
                         showItemSummary={this.state.showItemSummary}
                         tooltipContent={this.state.tooltipContent}
+                        setOverride={this.setOverride.bind(this)}
```
pinzart90, line 394: what is this used for ?

src/components/LibraryContainer.tsx
```diff
@@ -72,12 +75,13 @@ export class LibraryContainer extends React.Component<LibraryContainerProps, Lib
...
         this.scrollToExpandedItem = this.scrollToExpandedItem.bind(this);
         this.setTooltipText = this.setTooltipText.bind(this);
+        this.setHostContext = this.setHostContext.bind(this);
 
         // Set handlers after methods are bound.
         this.props.libraryController.setLoadedTypesJsonHandler = this.setLoadedTypesJson;
         this.props.libraryController.setLayoutSpecsJsonHandler = this.setLayoutSpecsJson;
         this.props.libraryController.refreshLibraryViewHandler = this.refreshLibraryView;
-
+        this.props.libraryController.setHostContextHandler = this.setHostContext;
```
pinzart90, line 84: I don't see these methods (setHostContext) called anywhere ? are they called on the c#/webview2 side ?

pinzart90: I see it now.

src/components/SearchResultItem.tsx
mjkkirschner: that's a good question, I can probably just check for null further up and bail.

src/components/LibraryContainer.tsx
mjkkirschner: yes, see the dynamo pr.
https://github.com/DynamoDS/Dynamo/pull/15204

src/components/LibraryContainer.tsx
mjkkirschner: see the description:
`
LibraryContainer.State.shouldOverideExpandedState  also controls the rendering behavior of LibraryItems - I found that LibraryItems today use a legacy life cycle hook UNSAFE_onWillRecieveProps - to override state with props - specifically for the expanded state so that when interacting only one library item can be open at a time.
`

TLDR: it's used to set a specific rendering mode where props.expanded overrides state.expanded of library items so they close.

mjkkirschner: ah, but your point is that this function setting the state is never called, thanks, I'll get rid of it, I ended up needing to set the state manually to avoid triggering a render, not ideal, but I think acceptable for now, I will get rid of this method.

pinzart90 (approved):
COuple of question, then LGTM

src/components/LibraryContainer.tsx
mjkkirschner: this is still necessary, though I moved it out of the `updateSearchViewDelayed` method - it's just a bug that only exists in the librarieJS test app, not in the dynamo context - see the comments I added in latest commit.

src/sharedTypes.ts
mjkkirschner: @pinzart90 - I figured this out, using string enums in newer versions of typescript works well and is easy to use from js as well.

PR #229, opened 2024-05-14
Feature/prepare librariejs bundle and resources

Enzo707:
- Move js implementations from index.html to main.js file
- Fix webpack config including all the resources in build package
- Launch build version (/dist) locally using express 

Review
@QilongTang 

FYI
@mjkkirschner


Enzo707:
Ready for review @QilongTang PTAL, thanks

QilongTang:
Do you think these changes would affect the layout spec for our integrators? e.g. https://github.com/DynamoDS/DynamoRevit/blob/master/src/DynamoRevit/Resources/LayoutSpecs.json

Enzo707:
> Do you think these changes would affect the layout spec for our integrators? e.g. https://github.com/DynamoDS/DynamoRevit/blob/master/src/DynamoRevit/Resources/LayoutSpecs.json


@QilongTang  I don't think it should but might be wise to ask others with more background than I to take a look just to be sure

mjkkirschner:
I am pretty sure that the layout spec in this repo (and of course the loaded types), should not be copied over to Dynamo. They are not accurate to what is in Dynamo.

The layout spec and types in this repo are loaded just for the debug app that is served by the express server.

The Revit layout spec is applied on top of the layout spec in Dynamo.

I'm not sure how your changes might affect that but TLDR - don't copy the layout spec or loaded types from this repo over the ones in Dynamo when it gets integrated there.

src/components/LibraryItem.tsx
```diff
@@ -167,7 +167,7 @@ export class LibraryItem extends React.Component<LibraryItemProps, LibraryItemSt
                 <div className="LibraryItemIconWrapper">
                     <img
                         className={"LibraryItemIcon"}
-                        src={this.props.data.iconUrl}
+                        src={require("../"+this.props.data.iconUrl)}
```
mjkkirschner, line 170: so in the context of a browser,  and not node, what does this do when transpiled?

Enzo707: > so in the context of a browser, and not node, what does this do when transpiled? Maybe another way of asking this question - why is this required?

This allows webpack to get the resource and include it into the build

mjkkirschner: How does web pack include these images in the build? I think you might find that on the Dynamo side we do some odd things to load/embed images as base64 data.

Enzo707: librariejs generates the build with webpack including the bundle, index.html and the resources which are served using express for debug proposes. 
On Dynamo side we should load the build as we do in NotificationsCenter but as you mention there are odd things to handle resources. I'm debugging LibraryViewController now, and seems like we replace the resources url in the bundle according to Dynamo's current LayoutSpects.json

QilongTang: Thank you for the explanation!

QilongTang (approved):

QilongTang:
Hi @Enzo707 Should we close this PR after our discussions?

PR #227, opened 2024-05-07
Fixing DevTools Console Errors

RobertGlobant20:
This fix will remove most of errors shown in the Browser-DevTools console for **Chrome** when the library is loaded (problem reported by @mjkkirschner ).
image

The one related to LIBPLACEHOLDER will be still present (a similar to this one is still present in all the React components that we have in Dynamo).

FYI: @QilongTang @reddyashish

PR #231, opened 2024-05-16
[DYN 2316] consume library as npm package

Enzo707:
This PR includes:

- Revert [PR #227 ](https://github.com/DynamoDS/librarie.js/pull/227) 
- Revert [PR #225](https://github.com/DynamoDS/librarie.js/pull/225)

We keep js code implemented in Dynamo side  (**LibraryViewExtensionWebView2/web/library/library.html**) as it is not relevant for other integrators. So these PRs are not needed anymore.
 
 Add command line executed on `npm run production` to remove resources folder as it's not necessary for integrators. Therefore we'll just publish librarie.js bundle but keeping the resources path in the bundle so they can be overwrite on Dynamo side by Dynamo's own resources.

**REVIEW:**
@QilongTang 
@RobertGlobant20 

**FIY**
@mjkkirschner

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,10 +19,11 @@ module.exports = {
     target: "web",
     output: {
         filename: productionBuild ? "librarie.min.js" : "librarie.js",
-        path: path.join(__dirname, '/dist/build'),
-        publicPath: "./dist/build/",
+        path: path.join(__dirname, '/dist/'),
+        publicPath: "./dist/",
         libraryTarget: "umd",
         library: "LibraryEntryPoint",
+        clean:  true
     },
     optimization:{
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -19,10 +19,11 @@ module.exports = {
     target: "web",
     output: {
         filename: productionBuild ? "librarie.min.js" : "librarie.js",
-        path: path.join(__dirname, '/dist/build'),
-        publicPath: "./dist/build/",
+        path: path.join(__dirname, '/dist/'),
```
QilongTang, line 22: Can we preserve the build folder? This would be consistent than other web components

Enzo707: Done!

